PR #40, opened 2025-08-29
Update README example to make OPA explicitly read files as bundles

johanfylling:
Some OPA versions have a halting [bug](https://github.com/open-policy-agent/opa/issues/7870) when watching bundles and bundle-mode isn't explicit. Using the `-b` flag avoids a panic on bundle reload.


sspaink:
@johanfylling the failing tests are being ignored for now: https://github.com/open-policy-agent/opa-control-plane/commit/48d9be8f00218e9fd05020105b437b3526db71e2

A rebase should give you all passing checks ✅

PR #41, opened 2025-09-02
docs: Prepare for website deployment

charlieegan3:
There will need to be more changes in future, but this is all that's needed for an initial pass.

PR #42, opened 2025-09-03
docs: Prepare for website deployment

charlieegan3:
Replaces https://github.com/open-policy-agent/opa-control-plane/pull/41

https://github.com/open-policy-agent/opa/pull/7875 will bring these docs into the main OPA repo to be managed there going forward.

There will be some future readme updates too, but for now I've mostly left in place, but we don't really want to have this content duplicated in the long run.


johanfylling (approved):
LGTM 👍

PR #51, opened 2025-09-09
ci: run golangci-lint

sspaink:


philipaconrad:
@sspaink This looks great, but the massive number of linter violations means we'll probably need to have a fat "linter cleanups" commit paired with the actual golangci-lint CI setup.

internal/pool/pool.go
```diff
@@ -93,7 +93,7 @@ func (p *Pool) dequeue() *task {
 			p.mu.Unlock()
 
 			select {
-			case <-time.After(t.deadline.Sub(time.Now())):
+			case <-time.After(time.Until(t.deadline)):
```
philipaconrad, line 96: I remember the `time.After` idiom being an issue in earlier Golang versions. As long as we're Go 1.23+, this won't be a resource leak. ([`time.After` docs](https://cs.opensource.google/go/go/+/go1.25.1:src/time/sleep.go;l=202))

internal/server/server.go
```diff
@@ -442,7 +442,7 @@ func getBoolParam(url *url.URL, name string, ifEmpty bool) bool {
 	}
 
 	for _, x := range p {
-		if strings.ToLower(x) == "true" {
+		if strings.EqualFold(x, "true") {
```
philipaconrad, line 445: TIL about [`strings.EqualFold`](https://cs.opensource.google/go/go/+/go1.25.1:src/strings/strings.go;l=1194), which does case-insensitive string compare for unicode! :slightly_smiling_face:

philipaconrad (approved):
This looks great! Thanks for mopping up all of those small linter nits.

For others looking at this changeset, most of the linter cleanups boil down to:
 - Extensive slice preallocation! :smile:
 - `fmt.Errorf` -> `errors.New`
 - Modern loop range syntax
 - Ignore assignments for errors we don't care about
 - Explicit linter check disables, when deprecated imports are still needed

Some review comments follow, but they're mostly remarks on what I saw while trawling through the changes:

cmd/backtest/backtest.go
```diff
@@ -21,9 +20,9 @@ import (
 
 	"github.com/akedrou/textdiff"
 	"github.com/olekukonko/tablewriter"
-	"github.com/open-policy-agent/opa/ast"
-	"github.com/open-policy-agent/opa/bundle"
-	"github.com/open-policy-agent/opa/sdk"
+	"github.com/open-policy-agent/opa/ast"    // nolint:staticcheck
```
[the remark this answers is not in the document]
sspaink: Correct!

PR #63, opened 2025-09-12
 internal/gitsync: If configuration changes, wipe the clone.

koponen:
Cloned repository holds the origin information and reconfiguring the URL should trigger re-cloning. However, instead of blindly wiping the cloned repositories during boot or reconfiguration, persist the configuration to filesystem and detect the configuration changes to avoid unnecessary wiping, and therefore, cloning.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -22,7 +22,7 @@ Example:
 ```yaml
 tokens:
     admin:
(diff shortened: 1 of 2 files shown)


srenatus (approved):
I have yet to get a feel for how things are done over here -- but would you think adding a test is worthwhile?

PR #71, opened 2025-09-17
server: include source, bundle, stack name in API results

### LINKED ISSUE #67: Add source identifier to v1/sources endpoint
When calling GET v1/sources the response only gives a list of gitsettings without any identifiers
We want to expand this endpoint by adding a "name" field next to gitsettings describing name of source

srenatus:
Previously, GET /v1/sources would give you a list of objects, corresponding to the sources; but it would omit their name. Now, it'll be included:

```
$ curl http://127.0.0.1:8282/v1/sources -H "Authorization: Bearer admin-api-key-change-me" | jq                                                                                           
{
  "result": [
    {
      "git": {
        "repo": "https://github.com/xxx/old-repo",
        "reference": "refs/heads/master",
        "path": "policy",
        "included_files": [
          "*.rego"
        ],
        "excluded_files": [
          "*.test.rego"
        ]
      },
      "name": "source-name"
    },
    {
      "git": {
        "repo": "https://github.com/xxx/old-repo",
        "reference": "refs/heads/master",
        "path": "policy",
        "included_files": [
          "*.rego"
        ],
        "excluded_files": [
          "*.test.rego"
        ]
      },
      "name": "source-two"
    }
  ]
}
````

Fixes #67.

Also applies the same change (plus added tests) to GET /v1/bundles an GET /v1/stacks.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -438,10 +438,8 @@ func (s *Source) SetEmbeddedFiles(files map[string]string) {
 }
 
 func (s *Source) SetPath(path string) {
-	for _, p := range s.Paths {
-		if p == path {
-			return
-		}
+	if slices.Contains(s.Paths, path) {
+		return
 	}
 
 	s.Paths = append(s.Paths, path)
@@ -459,7 +457,7 @@ func (a Sources) Equal(b Sources) bool {
 
 // Stack defines the configuration for a OPA Control Plane Stack.
 type Stack struct {
-	Name            string       `json:"-" yaml:"-"`
+	Name            string       `json:"name" yaml:"-"`
 	Selector        Selector     `json:"selector" yaml:"selector"` // Schema validation overrides Selector to object of string array values.
 	ExcludeSelector *Selector    `json:"exclude_selector,omitempty" yaml:"exclude_selector,omitempty"`
 	Requirements    Requirements `json:"requirements,omitempty" yaml:"requirements,omitempty"`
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by srenatus** [comment]
I'll take care of bundles, too. And perhaps more. 🔍

File: internal/config/config.go [hunk 1]
```diff
@@ -232,7 +232,7 @@ func (r *Root) Validate() error {
 
 // Bundle defines the configuration for a OPA Control Plane Bundle.
 type Bundle struct {
-	Name          string        `json:"-" yaml:"-"`
```
philipaconrad, line 235: The biggest question I have, is if omitting the `name` field was *intentional* for some reason? Maybe @koponen might know? 🤔

internal/config/config.go
```diff
@@ -438,10 +438,8 @@ func (s *Source) SetEmbeddedFiles(files map[string]string) {
 }
 
 func (s *Source) SetPath(path string) {
-	for _, p := range s.Paths {
-		if p == path {
-			return
-		}
+	if slices.Contains(s.Paths, path) {
```
philipaconrad, line 441: Nice modernization here. 🙂

philipaconrad (approved):
This looks like a small but solid upgrade in functionality to me. I've got a question about if the omission of the `name` field was intentional, which should show up in the review comments below.

Aside from the "headliner" changes in this PR, the test suite in `server_test.go` now does much more thorough checking/reporting of diffs from the expected return values, which is great! 🙂 Thanks for the upgrades there, @srenatus.

srenatus:
> I suppose this will also need an update to OCP's [API reference](https://www.openpolicyagent.org/docs/ocp/api-reference)?

Yes! Thanks for the reminder, I wasn't aware of those docs. Will update in the PR.

srenatus:
Oh actually those docs are not from this repo, are they? I'll find them and update them where they are.

srenatus:
👉 https://github.com/open-policy-agent/opa/pull/7906 Here we go 📚

PR #72, opened 2025-09-17
Dropping `go-test` timeout

johanfylling:
To not panic on slower systems, and to give us some time before GH turns flaky ;).


srenatus:
But I still wonder why the mysql tests don't work for you 🤔

johanfylling:
They do _work_; just not fast enough to fit inside 90s for some reason ..

Issue #3, opened 2025-05-02
Publish Docker image


tsandall:
This is self-explanatory. We'll want to distribute OCP as a Docker image.

PR #74, opened 2025-09-18
ci: Add Makefile targets and update GHA workflow to publish images

ashutosh-narkar:
This change adds the Makefile targets that create OCP Docker images and pushes them to Docker Hub. It also updates the GHA workflow such that the edge variant of the image containing the latest changes is pushed to Docker Hub on pushes to main.

Fixes: #3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ bundles/
 data
 opactl_*
 files
+.go
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,16 @@
 CGO_ENABLED ?= 0
 GOFLAGS ?= "-buildmode=exe"
 GO := CGO_ENABLED=$(CGO_ENABLED) GOFLAGS="$(GOFLAGS)" go
+
+VERSION := $(shell ./build/get-build-version.sh)
+
 DOCKER := docker
+
+DOCKER_UID ?= 0
+DOCKER_GID ?= 0
+
+REPOSITORY ?= openpolicyagent
+IMAGE := $(REPOSITORY)/opa-control-plane
 
 GOARCH := $(shell go env GOARCH)
 GOOS := $(shell go env GOOS)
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -12,3 +12,4 @@ bundles/
 data
 opactl_*
 files
+.go
```
srenatus, line 15: Hm? Where would that file come from? 🤔

Makefile
```diff
@@ -62,6 +136,23 @@ library-test:
...
 
+
+CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
+        $(DOCKER_FLAGS) \
+        -u $(DOCKER_UID):$(DOCKER_GID) \
+        -v $(PWD):/src \
+        -w /src \
+        -e GOCACHE=/src/.go/cache \
+        -e CGO_ENABLED=$(CGO_ENABLED) \
+        -e WASM_ENABLED=$(WASM_ENABLED) \
```
srenatus, line 147: [nit] We don't need that here, I believe.

Makefile
```diff
@@ -30,6 +54,10 @@ generate:
 .PHONY: build
 build: go-build
 
+.PHONY: build-linux-static
```
srenatus, line 57: [nit] I think I'd rather drop the `-static`. Let's not repeat past mistakes -- there's one binary for each platform, let's try to keep it at that 😅

srenatus (approved):
LGTM, a bunch of nitpicks, but nothing that would require much work to fix 😎 Thanks! 🎈

.gitignore
ashutosh-narkar: https://github.com/open-policy-agent/opa-control-plane/pull/74/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R145

Makefile
ashutosh-narkar: Ok removed.

Issue #66, opened 2025-09-15
'requirements' field for a sources is add-only


HookFirebolt:
Like reported in #53, the `PUT /v1/sources` endpoint is also add-only. This does not affect us so we haven't looked into the documentation to see if it makes sense to not make the same changes that we made to /v1/bundles to /v1/sources.

If they should be merged in the same pull request, let me know before merging #65, and I will add it.

johanfylling:
Are you referring to a specific field when you say the `/v1/sources` endpoint is add-only?

e.g.:

Creating a source:

```
PUT http://localhost:8282/v1/sources/gh

{
    "git": {
        "repo": "https://github.com/my-org/my-repo.git",
        "reference": "refs/heads/main",
        "path": "src",
        "included_files": ["*.rego"],
        "credentials": "github-readonly"
    }
}
```

Getting the source:

```
GET http://localhost:8282/v1/sources/gh

->

{
  "result": {
    "git": {
      "repo": "https://github.com/my-org/my-repo.git",
      "reference": "refs/heads/main",
      "path": "src",
      "included_files": [
        "*.rego"
      ]
    }
  }
}
```

Updating the `included_files` field of the same source:

```
PUT http://localhost:8282/v1/sources/gh

{
    "git": {
        "repo": "https://github.com/my-org/my-repo.git",
        "reference": "refs/heads/main",
        "path": "src",
        "included_files": ["*.rego", "*.json"],
        "credentials": "github-readonly"
    }
}
```

Refetching the source:

```
GET http://localhost:8282/v1/sources/gh

->

{
  "result": {
    "git": {
      "repo": "https://github.com/johanfylling/rego-utils.git",
      "reference": "refs/heads/main",
      "path": "src",
      "included_files": [
        "*.rego",
        "*.json"
      ]
    }
  }
}
```

Notice the updated `included_files`.

---

There seems to be an issue with the API not properly returning the data for "directory" type sources, which we're looking into.

PR #76, opened 2025-09-18
internal/database: delete requirements missing in PUT sources call

johanfylling:
Fixes: #66

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -1164,6 +1164,8 @@ func (d *Database) UpsertSource(ctx context.Context, principal string, source *c
 			}
 		}
 
+		// Upsert data sources
+
 		for _, datasource := range source.Datasources {
 			bs, err := json.Marshal(datasource.Config)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


internal/database/database.go
```diff
@@ -1188,11 +1192,21 @@ func (d *Database) UpsertSource(ctx context.Context, principal string, source *c
...
 					return err
 				}
+				sources = append(sources, *r.Source)
+			}
+		}
+
+		if source.Requirements != nil {
+			if err := d.deleteNotIn(ctx, tx, "sources_requirements", "source_name", source.Name, "requirement_name", sources); err != nil {
+				return err
 			}
```
srenatus, line 1210: This looks decent. What I started wondering is if a plain "delete from sources_requirements where source_name = ...." followed by an insert of the current ones wouldn't be simpler. Well, it most likely _is_ simpler, but I guess since it's two calls either way.... potato, potato.

srenatus (approved):
Thank you!

PR #77, opened 2025-09-18
ci: Add trust for mounted dir in container & push images on pushes to…

ashutosh-narkar:
… main only

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,12 +145,11 @@ CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
         -e GOCACHE=/src/.go/cache \
         -e CGO_ENABLED=$(CGO_ENABLED) \
 		-e GOARCH=$(GOARCH) \
-        golang:$(GOVERSION) \
-		make
+        golang:$(GOVERSION)
 
 .PHONY: ci-go-%
 ci-go-%:
-	$(CI_GOLANG_DOCKER_MAKE) "$*"
+	$(CI_GOLANG_DOCKER_MAKE) /bin/bash -c "git config --global --add safe.directory /src && make $*"
 
 .PHONY: clean
 clean:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -31,6 +31,7 @@ jobs:
 
   deploy-edge:
     name: Push Edge Release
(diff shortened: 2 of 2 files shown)


philipaconrad:
Another option to consider is isolating the "only on pushes to `main`" job into a separate YAML file, like how OPA/EOPA do it. You could then have it only run on pushes to main, just by the trigger controls at the top of the file, like so:
```yaml
name: Your Workflow Name Here

on:
  workflow_dispatch: {}
  push:
    branches: [main]

jobs:
  publish_job:
    ...
  ...
```

Makefile
```diff
@@ -145,12 +145,11 @@ CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
...
         -e CGO_ENABLED=$(CGO_ENABLED) \
 		-e GOARCH=$(GOARCH) \
-        golang:$(GOVERSION) \
-		make
+        golang:$(GOVERSION)
 
 .PHONY: ci-go-%
 ci-go-%:
-	$(CI_GOLANG_DOCKER_MAKE) "$*"
+	$(CI_GOLANG_DOCKER_MAKE) /bin/bash -c "git config --global --add safe.directory /src && make $*"
```
philipaconrad, line 152:
```suggestion
	/bin/bash -c "git config --global --add safe.directory /src && $(CI_GOLANG_DOCKER_MAKE) $*"
```
This way, we avoid accidentally invoking the *user's* `make` (or causing an error if it's not available, like on some Windows systems).

ashutosh-narkar: I'm not sure I understand. The `/src` directory in mounted inside the container and hence we run the command within the container. Maybe I'm missing something. [This](https://github.com/open-policy-agent/opa/blob/main/Makefile#L269) is what I referred to.

ashutosh-narkar:
> Another option to consider is isolating the "only on pushes to `main`" job into a separate YAML file, like how OPA/EOPA do it. You could then have it only run on pushes to main, just by the trigger controls at the top of the file, like so:
> 
> ```yaml
> name: Your Workflow Name Here
> 
> on:
>   workflow_dispatch: {}
>   push:
>     branches: [main]
> 
> jobs:
>   publish_job:
>     ...
>   ...
> ```

Sure. I was considering that option. But opa-envoy-plugin uses a similar approach so just went it instead of creating a new flow.

PR #83, opened 2025-09-19
CI: add test case for git ref with ssh-auth

srenatus:
Using a passphrase-protected deploy key with github.com/open-policy-agent/contrib.

The repo is OSS, but I've put the deploy key into the action secrets anyways. It'll raise less eyebrows if there's no "BEGIN PRIVATE KEY" in the txtar.


johanfylling (approved):
👍

PR #86, opened 2025-09-19
config|database: make 'datasources->credentials' work

srenatus:
Before, this wasn't persisted to the database, and not read back. So referencing a token among the secrets (type "token_auth") had no effect.

-----

So I have manually verified that the API works with this, too -- but I have yet to add some more tests for that.

Also, the added txtar case works -- but it also worked when the credentials didn't work: the GH API just uses different rate limiting when the header is present.

### DIFF
--- a/e2e/cli/build_http_transform.txtar
+++ b/e2e/cli/build_http_transform.txtar
@@ -1,0 +1,45 @@
+exec $OPACTL build --config config.d/bundle.yml --data-dir tmp --non-interactive --log-level debug
+stderr '^1/1 bundles built and pushed successfully$'
+! stdout .
+
+exec tar tf bundles/hello-world/bundle.tar.gz
+cmp stdout exp/tarball
+
+exec tar xf bundles/hello-world/bundle.tar.gz /data.json
+cmp data.json exp/data.json
+
+-- config.d/bundle.yml --
+bundles:
+  hello-world:
+    object_storage:
+      filesystem:
+        path: bundles/hello-world/bundle.tar.gz
+    requirements:
+    - source: http-with-transform
+    excluded_files:
+    - transform.rego
+sources:
+  http-with-transform:
+    datasources:
+    - type: http
+      name: hello
+      path: github
+      transform_query: data.transform.rule
+      config:
+        url: https://api.github.com/repos/open-policy-agent/opa/releases/tags/v1.8.0
+      credentials: github-token # NOTE(sr): There's no easy way to tell that this had an effect
+    directory: aux/
+    paths:
+    - transform.rego
+secrets:
+  github-token:
+    type: token_auth
+    token: ${E2E_GITHUB_TOKEN}
+-- aux/transform.rego --
+package transform
+rule.created := input.created_at
+-- exp/tarball --
+/data.json
+/.manifest
+-- exp/data.json --
+{"github":{"created":"2025-08-28T14:49:47Z"}}
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -826,9 +826,13 @@ WHERE (sources_secrets.ref_type = 'git_credentials' OR sources_secrets.ref_type 
 		sources_datasources.path,
 		sources_datasources.type,
 		sources_datasources.config,
-		sources_datasources.transform_query
+		sources_datasources.transform_query,
+	    sources_datasources.secret_name,
+	    secrets.value AS secret_value
 	FROM
 		sources_datasources
+	LEFT JOIN
+		secrets ON sources_datasources.secret_name = secrets.name
 	`)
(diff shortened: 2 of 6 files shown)


internal/database/schema.go
```diff
@@ -94,11 +87,13 @@ var schema = []sqlTable{
 	createSQLTable("sources_datasources").
 		VarCharNonNullColumn("name").
 		VarCharNonNullColumn("source_name").
+		TextColumn("secret_name").
```
srenatus, line 90: Decided against another cross-table, when we have all we need here already.

internal/httpsync/httpsync.go
```diff
@@ -81,7 +81,8 @@ func (s *HttpDataSynchronizer) setHeaders(ctx context.Context, req *http.Request
 		req.SetBasicAuth(value.Username, value.Password)
 	case config.SecretTokenAuth:
 		req.Header.Set("Authorization", "Bearer "+value.Token)
+	default:
+		return fmt.Errorf("unsupported authentication type: %T", value)
 	}
-
-	return fmt.Errorf("unsupported authentication type: %T", value)
+	return nil
```
srenatus, line 87: Fun bug! After the credential was pulled out of the database properly, it ended up failing here unconditionally 😅

internal/database/schema.go
```diff
@@ -78,13 +78,6 @@ var schema = []sqlTable{
 		PrimaryKey("source_name", "secret_name").
 		ForeignKey("source_name", "sources(name)").
 		ForeignKey("secret_name", "secrets(name)"),
-	createSQLTable("sources_secrets").
-		VarCharNonNullColumn("source_name").
-		VarCharNonNullColumn("secret_name").
-		TextNonNullColumn("ref_type").
-		PrimaryKey("source_name", "secret_name").
-		ForeignKey("source_name", "sources(name)").
-		ForeignKey("secret_name", "secrets(name)"),
```
srenatus, line 87: This was redundant. The same lines are above it.

internal/database/schema.go
```diff
@@ -94,11 +87,13 @@ var schema = []sqlTable{
 	createSQLTable("sources_datasources").
 		VarCharNonNullColumn("name").
 		VarCharNonNullColumn("source_name").
+		VarCharColumn("secret_name").
```
srenatus, line 90: MySQL won't let us use `TextColumn` with `ForeignKey`, because it builds an index for those and cannot put an index on TEXT without a "prefix count", which basically is the `VarChar(255)` we'll put there for `VarCharColumn`. Postgres and SQLite didn't care.

internal/database/database.go
```diff
@@ -1172,9 +1185,13 @@ func (d *Database) UpsertSource(ctx context.Context, principal string, source *c
 				return err
 			}
 
-			if err := d.upsert(ctx, tx, "sources_datasources", []string{"source_name", "name", "type", "path", "config", "transform_query"},
+			var secret sql.NullString
+			if datasource.Credentials != nil {
+				secret.String, secret.Valid = datasource.Credentials.Name, true
```
johanfylling, line 1190: Funny there is no helper factory function for creating a `sql.NullString` that correctly assigns the `Valid` field. Suppose this type is mostly for reading data(?).

internal/service/service.go
```diff
@@ -298,9 +298,9 @@ func (s *Service) launchWorkers(ctx context.Context) {
 			sort.Strings(sorted)
 			var extra string
 			if len(sorted) > 1 {
-				extra = " (along with %d other sources)"
+				extra = fmt.Sprintf(" (along with %d other sources)", len(sorted)-1)
```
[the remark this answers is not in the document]
johanfylling: So `len(sorted)` was never `> 1`; or we never checked the result?

internal/test/dbs/dbs.go
```diff
@@ -25,6 +27,20 @@ func tcCleanup(t *testing.T, ctr testcontainers.Container) func() {
 	}
 }
 
+// This DSN is for separating multiple in-mem SQLite instances. Each test will get
+// it's own in-memory "file".
```
johanfylling, line 31: 👍

e2e/cli/build_http_transform.txtar
```diff
@@ -0,0 +1,64 @@
...
+  api-token:
+    type: token_auth
+    token: open sesame
+  basic-auth:
+    type: basic_auth
+    username: user001
+    password: letmein
+-- aux/transform.rego --
+package transform
+import rego.v1
```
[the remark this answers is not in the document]
srenatus: We should make it so, I think. It's needed at the moment, conservative settings for DAS migration, I suppose.

PR #93, opened 2025-09-23
Add DELETE handlers for sources, stacks, bundles

### LINKED ISSUE #17: Implement DELETE endpoints for v1/sources, v1/stacks/, and v1/bundles

srenatus:
I hope I got the hierarchies right. You can only delete a source if it's not referenced in a bundle, or a stack.

Fixes #17.

### DIFF
--- a/internal/database/database_test.go
+++ b/internal/database/database_test.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/styrainc/opa-control-plane/internal/config"
 	"github.com/styrainc/opa-control-plane/internal/database"
 	"github.com/styrainc/opa-control-plane/internal/service"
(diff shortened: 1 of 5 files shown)


internal/database/database_test.go
```diff
@@ -271,8 +289,13 @@ func TestDatabase(t *testing.T) {
 					Name:         "system6",
 					Requirements: config.Requirements{},
 				}),
+				newTestCase("delete bundle and sources"). // We check that it deletes cross-refs by attempting to delete those before and after the bundle delete: TODO(sr)
+										DeleteSource("source-a", false). // Cannot delete source, it's referenced
```
srenatus, line 293: > One thing that might be nice to have is a test of the "can't delete a source if it's referenced in a bundle or a stack" behavior.

This is it. Deleting "source-a" fails at first because it's references in "bundle-a". After that's deleted, we'll be able to successfully delete "source-a".


I'll add a case for stack -> source, good point.

Issue #92, opened 2025-09-23
gorilla/mux -> net/http mux


srenatus:
There is no need to use gorilla here, anymore, I think. Let's switch to the stdlib [ServeMux](https://pkg.go.dev/net/http#ServeMux).

srenatus:
@philipaconrad Thanks for pushing this through so quickly. Nice surprise this morning 🎉

PR #95, opened 2025-09-23
refactor: Swap out gorilla/mux -> net/http ServeMux

philipaconrad:
## What changed?

This PR migrates the server from using `gorilla/mux` to using the stdlib's `net/http` [`ServeMux` type](https://pkg.go.dev/net/http#ServeMux). A few minor changes were needed in the handlers to extract the path wildcard vars.

Resolves: #92


srenatus (approved):
Thanks a lot! Just one nitpick

PR #98, opened 2025-09-25
Fix failing cli builds with --non-interactive flag exiting with success

selaux:
os.Exit(1) was only called within `printReport` which is only executed when `--non-interactive` is not used. This lead to failing bundle builds still resulting in exit code 0 when the flag was used.

After this commit the build results are collected regardless of the `--non-interactive` flag and the exit code is set properly based on those results.

### DIFF
--- a/cmd/build/build.go
+++ b/cmd/build/build.go
@@ -85,8 +85,14 @@ func init() {
 				os.Exit(1)
 			}
 
+			buildReport := collectBuildReport(svc.Report())
+
 			if !params.noninteractive {
-				printReport(svc.Report())
+				printReport(buildReport)
+			}
+
+			if !buildReport.allBuildsSuccessful() {
+				os.Exit(1)
 			}
 		},
 	}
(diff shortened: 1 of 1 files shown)


srenatus:
Thank you. This is a good idea, you're of course correct about `--non-interactive` not making failing builds exit with code 0.

Would you mind adding a test? We've got this nice little [testscript setup here](https://github.com/open-policy-agent/opa-control-plane/tree/main/e2e/cli), and a script like
```
! exec $OPACTL build --config config.yaml

-- config.yaml --
some config that'll yield a problem
```

should do the trick, I think.

srenatus:
Heya!

>  I guess we can run two commands in the same test, right?

Yes, you can have any number of expressions in a test script. Currently, for `opactl build`, we have a one-thing-one-file situation, but that's really just coincidental. Having multiple `exec $OPACTL` calls (or `! exec $OPACTL` when errors are expected) in a single txtar is perfectly acceptable. :+1: LMK if you get stuck (I'm also on slack, of course).

PR #100, opened 2025-09-26
authz_test: make tests repeatable, fix flaky TestPartialStringArgs

srenatus:
The assertion in TestPartialStringArgs was checking a certain ordering of the args slice. The first part needs to match, but the second part doesn't. It came down to ordering of the PE result queries. We not check both (equally valid) versions.

This was shadowed by the tests running with the cache -- so multiple evals wouldn't _really evaluate_ the code that yielded the non-deterministic results.

### DIFF
--- a/internal/authz/authz_test.go
+++ b/internal/authz/authz_test.go
@@ -1,23 +1,28 @@
 package authz
 
 import (
-	"context"
 	"database/sql"
-	"fmt"
-	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	_ "modernc.org/sqlite"
+
+	"github.com/styrainc/opa-control-plane/internal/test/dbs"
 )
 
 func TestPartialStringArgs(t *testing.T) {
-	result, err := Partial(context.Background(), Access{Principal: "bob", Resource: "sources", Permission: "sources.view", Name: "x123"}, map[string]ColumnRef{"input.name": {Table: "sources", Column: "name"}})
+	// NOTE(sr): Don't use the cache here (Partial, uppercase 'P'), as it'll make running this
+	// test multiple times meaningless.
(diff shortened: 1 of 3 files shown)


internal/authz/authz_test.go
```diff
@@ -33,40 +36,32 @@ func TestPartialStringArgs(t *testing.T) {
 		"bob",
 		"viewer",
 	}
-	if !reflect.DeepEqual(expArgs, args) {
-		t.Fatalf("unexpected args\n\ngot: %v\n\nexp: %v", args, expArgs)
+	if diff := cmp.Diff(expArgs, args, cmpopts.SortSlices(func(a, b any) bool { return a.(string) < b.(string) })); diff != "" {
```
srenatus, line 39: Ordering of that slice is irrelevant for the filter. It seems from some indeterminate ordering in the PE result's queries, I didn't chase that down, because it had no semantic implications there, either.

internal/authz/authz_test.go
```diff
@@ -28,45 +33,45 @@ func TestPartialStringArgs(t *testing.T) {
...
-		"bob",
-		"viewer",
 	}
-	if !reflect.DeepEqual(expArgs, args) {
-		t.Fatalf("unexpected args\n\ngot: %v\n\nexp: %v", args, expArgs)
+	if diff := cmp.Diff(expArgs, args[:6]); diff != "" {
+		t.Fatal("unexpected first 6 args (-want, +got)", diff)
+	}
+
+	expBob1 := []any{
```
srenatus, line 41: Both versions are fine here. But we cannot know which one we get.

philipaconrad (approved):
These changes look good! Thanks for putting this PR together. I like the cleanups, and think the refactors make sense.

Also, the review comments were helpful for getting my bearings a little faster during review. 🙂

PR #101, opened 2025-09-26
http: expose secrets CRUD

### LINKED ISSUE #16: Implement v1/secrets

srenatus:
To support listing them, we're adding an auto-incrementing id column to the table.

Let's figure out some time in the future when we stop doing such moves, and introduce a migrations framework of some sort.

Also adding rows.Err() checks that are required by the driver, but had been missing in a few places.

Fixes #16 (when done)

TODOs
- [x] database_test
- [x] server handlers
- [x] server handler tests

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -476,7 +476,7 @@ func (s *Server) v1SecretsPut(w http.ResponseWriter, r *http.Request) {
 	if secret.Name == "" {
 		secret.Name = name
 	} else if secret.Name != name {
-		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, errors.New("stack name must match path"))
+		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, errors.New("secret name must match path"))
 		return
 	}
 


srenatus:
💭 I'm torn about List and Get for secrets. It feels like we should mask or hide the actual value. So you can update and delete the secrets, but you can never get a stored secret back out. @johanfylling what do you think?

internal/database/database.go
```diff
@@ -933,6 +942,102 @@ WHERE (sources_secrets.ref_type = 'git_credentials' OR sources_secrets.ref_type
...
+        secrets
+    WHERE (` + conditions + ")"
+
+		if opts.name != "" {
+			query += fmt.Sprintf(" AND (secrets.name = %s)", d.arg(len(args)))
+			args = append(args, opts.name)
+		}
+
+		if after := opts.cursor(); after > 0 {
+			query += fmt.Sprintf(" AND (secrets.id > %s)", d.arg(len(args)))
```
philipaconrad, line 997: [question]: Probably very out-of-scope for this PR, but... is there a way that we might be able to tap into [prepared statements](https://pkg.go.dev/database/sql#Tx.Stmt), instead of doing direct string substitution in the future? 🤔 

I'm fine creating a GH task issue for this, if we think it's something worth changing.

srenatus: The conditions come from PE and are created with each incoming request.... so it might be difficult? But, OTOH, the Rego never changes, so the number of possible statements is bounded, and we could therefore prepare them, couldn't we? (Probably has to be done lazily?)

philipaconrad: I guess my thought was that the string templating we're doing with `%s` could at least be replaced with something "stronger" in the future. The string chunks for the WHERE clause that's coming from PE should already be sanitized/safely-quoted, if memory serves.

philipaconrad (approved):
Looks good to me! I like the internal shift to using *references* (by name, with `*SecretRef`) to the secrets on the List endpoint, versus using `*Secret` directly.

That also makes it a little easier to ensure in the JSON response types that we don't accidentally cough up a secret value when we're not supposed to.

internal/server/server.go
```diff
@@ -403,6 +408,87 @@ func (s *Server) v1StacksPut(w http.ResponseWriter, r *http.Request) {
...
+	var secret config.Secret
+	if err := newJSONDecoder(r.Body).Decode(&secret); err != nil {
+		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, err)
+		return
+	}
+
+	if secret.Name == "" {
+		secret.Name = name
+	} else if secret.Name != name {
+		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, errors.New("stack name must match path"))
```
srenatus, line 479: c/p
```suggestion
		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, errors.New("secret name must match path"))
```

PR #108, opened 2025-09-29
database: handle database changes in migrations

### LINKED ISSUE #106: database migrations
So far, we've played fast and loose with our database schema. At some point, as soon as someone has data stored that can't be recreated easily, we'll need to handle our database schema with some more care: 

1. introduce some basic migrations setup
2. do changes to the schema with migrations
3. test this thoroughly

Let's collect common deployment modes -- especially if we (ever/already/want to) support multiple OCP with a single database. That'll inform how (1.) needs to work.

srenatus:
With this change, `opactl run` will do this:
1. if run without a database config (i.e. using the sqlite default), it'll apply migrations
2. otherwise, it'll check the current state of the migrations in the database, with "$latest" the latest migration compiled into the binary,
  a. it will warn about pending migrations  if $current < $latest
  b. it will warn about a stale binary if $current > $latest
  c. attempt to use the database as-is

Migrations can be applied out of band by the operations procedures of the user: `opactl db migrate` will do that. It's safe to run at any time, and should be the first thing you do after pulling a new OCP image.

For simplicity, when you deploy a single OCP that owns its database, you can pass `--apply-migrations` to `opactl run`. It'll then just apply the latest migrations as compiled into the binary. 

Fixes #106.

### DIFF
--- a/cmd/build/build.go
+++ b/cmd/build/build.go
@@ -14,6 +14,7 @@ import (
 	"github.com/styrainc/opa-control-plane/cmd/internal/flags"
 	"github.com/styrainc/opa-control-plane/internal/config"
 	"github.com/styrainc/opa-control-plane/internal/logging"
+	"github.com/styrainc/opa-control-plane/internal/migrations"
 	"github.com/styrainc/opa-control-plane/internal/progress"
 	"github.com/styrainc/opa-control-plane/internal/service"
 	"github.com/styrainc/opa-control-plane/internal/util"
@@ -26,6 +27,7 @@ type buildParams struct {
 	persistenceDir    string
 	resetPersistence  bool
 	mergeConflictFail bool
+	migrateDB         bool
 	bundleNames       []string
 	logging           logging.Config
 }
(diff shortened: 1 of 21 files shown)


srenatus:
This needs a docs follow-up in the website docs:
1. explain the `opactl db migrate` command
2. explain the flag you can use for simple deployments (to automatically apply migrations if it's only a single OCP w/ a single DB)

srenatus:
Docs PR is here: https://github.com/open-policy-agent/opa/pull/7952

internal/server/server_test.go
```diff
@@ -16,6 +16,7 @@ import (
```
srenatus, line 16: This is the fold, nothing too interesting happens below. Test updates. And updating the service init, as mentioned above.

cmd/build/build.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/styrainc/opa-control-plane/cmd/internal/flags"
 	"github.com/styrainc/opa-control-plane/internal/config"
 	"github.com/styrainc/opa-control-plane/internal/logging"
+	"github.com/styrainc/opa-control-plane/internal/migrations"
```
philipaconrad, line 17: I can't believe I just noticed this, but apparently no one has adjusted the `go.mod` to be `github.com/open-policy-agent/opa-control-plane`, instead of the original `github.com/styrainc` prefix. 🤦

philipaconrad: I'll file a task issue for that-- it doesn't need to be fixed in this PR, since it's a logically distinct change.

philipaconrad: https://github.com/open-policy-agent/opa-control-plane/issues/119

internal/service/service.go
```diff
@@ -177,20 +194,17 @@ shutdown:
 		for _, w := range s.workers {
 			s.report.Bundles[w.bundleConfig.Name] = w.status
 		}
-		for b, status := range s.failures {
-			s.report.Bundles[b] = status
-		}
+		maps.Copy(s.report.Bundles, s.failures)
```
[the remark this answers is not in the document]
srenatus: gopls suggested it, and made it a push-of-a-button effort 😄  (Or rather "Space-a, Enter", in helix.)

PR #110, opened 2025-09-30
builder: apply excludes when checking bundle roots

### LINKED ISSUE #107: Excluded files from git still cause conflicting packages
Just ran into this issue with our pipeline yamls in our policy repos. Excluding them somehow does not affect package conflicts. Situation is:

- Two different git repositories / sources for policies
- Both git repositories contain a `pipeline.yaml` file on the top level
- These `pipeline.yaml` files are explicitly in the `excluded_files` of the git repository, so the excluded files looks something like this: `[".*/*", "pipeline.yaml"]`

Running `opactl build` in this case results in:

```
0/1 bundles built and pushed successfully
+--------+--------------+---------------------------------------------------------------------------+
| BUNDLE |    STATUS    |                                  MESSAGE                                  |
+--------+--------------+---------------------------------------------------------------------------+
| bundle | BUILD_FAILED | requirement "package2" contains conflicting package <illegal path "data"> |
|        |              | - package <illegal path "data"> from "package1"                           |
|        |              | - package .git from "package1"                                            |
|        |              | - package package1 from "package1"                                        |
+--------+--------------+---------------------------------------------------------------------------+
```

In my opinion this should not result in any errors, as the path is ignored. Also not sure what this `<illegal path "data">` is about. Happy for any feedback if this is somehow expected and an issue on my side.

Minimal reproduction sample is here: https://github.com/selaux/ocp-policies-reproducer

srenatus:
Previously, the per-source excludes hadn't been applied when walking data and policy files to figure out the roots contributions of a source.

Now, the same FilterFS that is used for building the bundle is used to determine its roots, yielding more consistent results.

Fixes #107.


selaux:
Looks good to me.

PR #123, opened 2025-10-02
cmd/migrate: only iterate over stacks matched by systems

M4nuF:
Adapt the migrate command to only consider stacks that are actually referenced as matching stacks by systems.

Skip stacks that are not referenced by systems to speed up the migration. In larger Styra DAS setups with lots of stacks this speeds up the migration considerably - especially when only migrating a single system (CLI flag --system-id).

### DIFF
--- a/cmd/internal/das/das.go
+++ b/cmd/internal/das/das.go
@@ -22,6 +22,7 @@ type V1System struct {
 	BundleRegistry struct {
 		ManualDeployment bool `json:"manual_deployment"`
 	} `json:"bundle_registry"`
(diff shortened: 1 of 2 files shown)


srenatus:
> Question to you: How do you want me to test this? I didn't want to introduce too much new mocking logic to be able to unit test the changes I did. But if that is preferred I can also add a unit test-case and generate the necessary mocks.

I think this is fine. Not like it would break anything for anyone, now that it's behind a flag. And I presume that since you have a _need_ for this, you'll quite naturally use the feature, and it'll work for that use case. 🤞 

Thanks again for contributing!

PR #127, opened 2025-10-06
Fix using migration config

### LINKED ISSUE #126: Error when building a bundle with builtin libraries
We have a DAS entz system that we ran the migrate command to pull into opa control plane.  Upon running opactl build with the latest version, we receive the following output:

```
loading configuration 100% [================================================================================] (88/88)
building and pushing bundles 100% [================================================================================] (1/1) 
0/1 bundles built and pushed successfully
+----------------+--------------+-------------------------------------------------------+
|     BUNDLE     |    STATUS    |                        MESSAGE                        |
+----------------+--------------+-------------------------------------------------------+
| my-entz-system | BUILD_FAILED | build: load error: 10 errors occurred during loading: |
|                |              | open 1.0-completions: file does not exist             |
|                |              | open 1.0-completions1: file does not exist            |
|                |              | open 1.0-conflicts: file does not exist               |
|                |              | open 1.0-conflicts1: file does not exist              |
|                |              | open 1.0-entrypoint-main: file does not exist         |
|                |              | open 1.0-entrypoint-main1: file does not exist        |
|                |              | open 1.0-library: file does not exist                 |
|                |              | open 1.0-library1: file does not exist                |
|                |              | open 1.0-transform: file does not exist               |
|                |              | open 1.0-transform1: file does not exist              |
+----------------+--------------+-------------------------------------------------------+
```

In our `config.d` folder we have the following files, they have been reduced to the minimum configuration needed to reproduce the issue.

builtin-sources.yaml
```yaml
metadata:
    exported_from: redacted
    exported_at: "2025-10-03T15:30:59Z"
sources:
    template.entitlements:1.0-completions:
        builtin: entitlements-v1/completions
    template.entitlements:1.0-conflicts:
        builtin: entitlements-v1/conflicts
    template.entitlements:1.0-entrypoint-main:
        builtin: entitlements-v1/main
    template.entitlements:1.0-library:
        builtin: entitlements-v1/library
    template.entitlements:1.0-transform:
        builtin: entitlements-v1/transform
```

my-entz-system.yaml
```yaml
metadata:
    exported_from: redacted
    exported_at: "2025-10-03T15:30:59Z"
bundles:
    my-entz-system:
        labels:
            system-id: redacted
            system-type: template.entitlements:1.0
        object_storage:
            filesystem:
                path: bundles/my-entz-system/bundle.tar.gz
        requirements:
            - source: my-entz-system
sources:
    my-entz-system:
        requirements:
            - source: template.entitlements:1.0-entrypoint-main
            - source: template.entitlements:1.0-completions
            - source: template.entitlements:1.0-conflicts
            - source: template.entitlements:1.0-transform
            - source: template.entitlements:1.0-library
```

This issue was introduced with this change: https://github.com/open-policy-agent/opa-control-plane/pull/96

I tested by pulling the previous commit, rebuilding opactl and running a build against that folder successfully, but that commit and all subsequent cause the build failure outlined here.

srenatus:
Fixes #126


internal/builder/builder.go
```diff
@@ -178,30 +179,47 @@ func (b *Builder) Build(ctx context.Context) error {
...
+			rs, err := getRegoAndJSONRoots(fs_)
+			if err != nil {
+				return fmt.Errorf("%v: %w", next.Name, err)
+			}
+			newRoots = append(newRoots, rs...)
+			return nil
+		}
+
+		for i, fs0 := range next.FSes {
+			if err := checkRoots(fs0); err != nil {
```
selaux, line 206: Do we need to check for `b.excluded` here as well? I guess for the purpose of `builtins` probably not, but if it is ever reused for some other purpose we might run into issues.

Could make sense to implement a `source.Filesystems` method that unifies `source.FSes` and `source.Dirs` and does the source-level filtering. Then the filtering for `b.excluded` could be a FilterFS on top of that.

selaux (commented):
Checked the behavior after this PR and the result looks fine to me. The only thing I noticed is that the resulting bundle will still have some `:` occurences in the the later segments of the paths. E.g. `template.entitlements#1.0-library/global/systemtypes/entitlements:1.0/library`. Not sure if that will cause any issues with Windows.

Other than that a few minor suggestions the rest LGTM.

internal/builder/builder.go
srenatus: Heh that's a nice refactoring idea. I'll give it a try!

srenatus:
> The only thing I noticed is that the resulting bundle will still have some : occurences in the the later segments of the paths. E.g. template.entitlements#1.0-library/global/systemtypes/entitlements:1.0/library. Not sure if that will cause any issues with Windows.

I don't know either. Let's be sure and run `opa inspect` on the bundles in CI.

selaux:
Thanks for looking into the comments. LGTM 🙂

PR #142, opened 2025-10-16
server: expose PATCH for sources/data API

### LINKED ISSUE #105: Feature Request: Allow PATCH on Push Datasource
# Problem Description

For "push" datasources, clients might know what they want to change but not have the full state of the entire data at hand. This is true when access assignments are pushed, in which case a very small patch operation can operate on fairly large data sources. 

# Proposed Solution

Add a new http method PATCH to the existing API and allow sending a JSON Patch in the body. An important caveat is that this should allow creating missing paths as especially for access assignments there might be shared data structures that are dynamic and traversed with overlapping patches. 

A library like https://github.com/evanphx/json-patch does support creating missing paths when "add" is applied. 

# Out of scope 

Applying even a small patch on a multi-MB datasource can be slow as the entire blob is read, changed and written back to the database. Future enhancements might be storing the patches themselves and then applying them in batch and keeping the patches as an audit trail with timestamps.

srenatus:
We're cautiously only allowing three operations: add/replace/remove.

All paths in patches are relative to the query path, so
```yaml
op: add
path: /foo
value: bar
```
sent via PATCH to `/v1/sources/{source}/data/a/b/c` would add "bar" as value of key `a/b/c/foo`.

Fixes #105.


internal/server/server.go
```diff
@@ -317,6 +319,41 @@ func (s *Server) v1SourcesDataPut(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	var patch types.SourcesPatchDataRequestV1
+	if err := newJSONDecoder(r.Body).Decode(&patch); err != nil {
+		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, err)
+		return
+	}
+
+	previous, _, err := s.db.SourcesDataGetRaw(ctx, name, path.Join(r.PathValue("path"), "data.json"), s.auth(r))
```
selaux: Should getting the data, applying the patch and saving it back be done in a transaction? Or ist that already automatted through `s.db`?

srenatus: Yup that's a good point. I'll try to move the patch application one layer down -- the database API does not expose transactions.

srenatus: Thanks for your thoughts. Let's roll with this. If we change it, the requirement would become laxer, which is always easier to introduce.

PR #145, opened 2025-10-20
config,server: add configuration option to set api prefix for server

irodzik:
Sometimes there is a need to listen on different path than root. Corporate policy or widely applied convention. Added configuration option will allow for modifying the prefix via config entry.
Eg:
```yaml
service:
    api_prefix: /my/prefix
```
will change path for example for health endpoint to `/my/prefix/health`.
This applies to all endpoints.

### DIFF
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -73,7 +73,7 @@ func init() {
 			}
 
 			go func() {
-				if err := server.New().WithDatabase(svc.Database()).WithReadiness(svc.Ready).Init().ListenAndServe(params.addr); err != nil {
+				if err := server.New().WithDatabase(svc.Database()).WithReadiness(svc.Ready).WithConfig(config.Service).Init().ListenAndServe(params.addr); err != nil {
 					log.Fatalf("failed to start server: %v", err)
 				}
 			}()
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -40,6 +40,7 @@ type Root struct {
 	Secrets  map[string]*Secret `json:"secrets,omitempty" yaml:"secrets,omitempty"` // Schema validation overrides Secret to object type.
 	Tokens   map[string]*Token  `json:"tokens,omitempty" yaml:"tokens,omitempty"`
 	Database *Database          `json:"database,omitempty" yaml:"database,omitempty"`
+	Service  *Service           `json:"service,omitempty" yaml:"service,omitempty"`
 }
 
 // SetSQLitePersistentByDefault sets the database configuration to use a SQLite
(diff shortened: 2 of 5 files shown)


cmd/run/run.go
```diff
@@ -86,6 +87,7 @@ func init() {
 
 	flags.AddConfig(run.Flags(), &params.configFile)
 	run.Flags().StringVarP(&params.addr, "addr", "a", defaultLocalAddr, "Set listening address of the server")
+	run.Flags().StringVarP(&params.contextPath, "context-path", "", "", "Set context path (eg. /my/path) that will define root path of the server")
```
srenatus: Hmm. I kind of think that a config value would be more befitting? If you agree, let's add something like [this Service section](https://github.com/open-policy-agent/opa-control-plane/pull/124/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR44) (from another PR) and add `api_prefix` or something like that to it. I'd avoid "context" as it's a bit vaguer than needed, I think.

srenatus (commented):
Thanks for contributing. I've added a couple of comments inline.

For tests, I think it would be OK to create a variant of [this test](https://github.com/open-policy-agent/opa-control-plane/blob/main/e2e/cli/run_basic.txtar), with an adapted config file, a different port, and the `curl` calls including the prefix. 

Thanks again!

internal/config/config.go
```diff
@@ -990,6 +991,12 @@ type AmazonRDS struct {
 	RootCertificates string `json:"root_certificates,omitempty" yaml:"root_certificates,omitempty"`
 }
 
+type Service struct {
+	// ApiPrefix prefixes all endpoints (including health and metrics) with its value. It is important to start with `/`.
```
srenatus: 💭 Is it really that important that it starts with '/'? We could easily make it start with that if it doesn't

irodzik: Well, we can either enforce it or add `/` ourselves. I personally prefer doing it explicitly than later dig into code to understand why it was added. But this is just my preference. What do you think?

internal/server/server.go
```diff
@@ -36,12 +37,14 @@ func (s *Server) Init() *Server {
...
-	s.router.HandleFunc("GET /health", s.health)
+	contextPath := s.contextPath
+
+	s.router.Handle(contextPath+"/metrics", metrics.Handler())
+	s.router.HandleFunc("GET "+contextPath+"/health", s.health)
 
 	base := chain.New(authenticationMiddleware(s.db))
 	setup := func(method, pattern string, hndl http.HandlerFunc) {
-		s.router.Handle(method+" "+pattern, append(base, metrics.InstrumentHandler(pattern)).ThenFunc(hndl))
+		s.router.Handle(method+" "+contextPath+pattern, append(base, metrics.InstrumentHandler(pattern)).ThenFunc(hndl))
```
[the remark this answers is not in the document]
irodzik: After some thinking time, I would vote for keeping all of them aligned.

internal/server/server.go
```diff
@@ -92,6 +99,18 @@ func (s *Server) ListenAndServe(addr string) error {
 	return http.ListenAndServe(addr, s.router)
 }
 
+func (s *Server) WithConfig(config *config.Root) *Server {
```
srenatus: ❓ How about this, what do you think? We'll only need the service bits.
```suggestion
func (s *Server) WithConfig(config *config.Service) *Server {
```

internal/server/server.go
```diff
@@ -36,12 +39,16 @@ func (s *Server) Init() *Server {
 		s.router = http.NewServeMux()
 	}
 
-	s.router.Handle("/metrics", metrics.Handler())
-	s.router.HandleFunc("GET /health", s.health)
+	apiPrefix := s.apiPrefix
+
+	s.log.Debugf("Api prefix set to '%s'", apiPrefix)
```
srenatus: If we add a debug log, I'd make it something that contains all the information we can provide. Something like _"Listening on <url>:<port>/<prefix>"_. What do you think? Alternatively, I'd drop the log completely.

srenatus (commented):
Last round of nitpicks -- thanks for bearing with me. This looks good, just two small things 👇

internal/server/server.go
irodzik: Ok, providing this info in one place is not perfect, because address is sitting in params and and prefix resides in config. So i will drop it for now.

PR #174, opened 2025-11-07
gitsync: don't fetch from remote with hash that was already fetched

srenatus:
Since a git hash cannot change, there's no point in bugging anyone about it ever again, once we have it in our local clone.

~🗒️ Need to figure out a way to test this without waiting for subsequent git fetches...~


stiidk:
🙌 🎉

PR #186, opened 2025-11-12
config: add more "no additional properties" markers

srenatus:
Haven't quite nailed this yet! Some improvements.

I figured since we're potentially causing some friction with the YAML configs, this is a good moment to switch to a better-supported library.

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -1,6 +1,8 @@
 {
+  "additionalProperties": false,
   "definitions": {
     "ConfigAmazonRDS": {
+      "additionalProperties": false,
       "properties": {
         "credentials": {
           "type": "string"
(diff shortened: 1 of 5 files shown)


internal/config/secret.go
```diff
@@ -65,12 +64,6 @@ func (s *Secret) Ref() *SecretRef {
 	return &SecretRef{Name: s.Name, value: s}
 }
 
-func (*Secret) PrepareJSONSchema(schema *jsonschema.Schema) error {
-	schema.Type = nil
-	schema.AddType(jsonschema.Object)
-	return nil
-}
```
srenatus: This had no effect on the generated schema

PR #187, opened 2025-11-13
bundle: auto-mount matching stacks (add opt-out option)

srenatus:
When a stack matches now, it'll be injected with a prefix of

    data.stacks.<stack_name>

You can opt out of this behaviour by providing bundle options:

```yaml
bundles:
  hello-world:
    options:
      no_default_stack_mount: true
```

Existing `prefix` settings on the stack's requirements are respected, so

```yaml
stacks:
  stack1:
    selector: ...
    requirements:
    - source: lib1
      prefix: imported_lib1
```

will cause lib1 to be mounted on `data.stacks.stack1.imported_lib1`.

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -83,6 +83,9 @@
         "object_storage": {
           "$ref": "#/definitions/ConfigObjectStorage"
         },
+        "options": {
+          "$ref": "#/definitions/ConfigOptions"
+        },
         "rebuild_interval": {
           "type": "string"
         },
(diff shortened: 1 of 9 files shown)


scott2449:
@srenatus Why is this the default behavior? I thought about opening an issue but thought maybe better to ask here first? This breaks kubernetes conflict logic (at least for styra migrants like myself): 

https://github.com/open-policy-agent/opa-control-plane/blob/main/libraries/kubernetes-v2/conflicts/library/v1/stacks/decision/v2/v2.rego#L41-L48

I had to add this flag to the dozens of k8s systems I had and it took us a couple weeks to track down why they weren't working. Turns out with the addition of the prefix was rendering that logic useless.

Def too late to worry about Styra migration logic, but should this or conflict logic be flipped? Documented better?

My biggest worry is this changing later somehow and my bundles falling apart 😄 I spoke with @johnsnow-styra and he also felt is was worth getting clarification since it means lots of previous backtesting is now invalid.

PR #202, opened 2025-11-20
introduce tenants in the database layer

srenatus:
This is preliminary work enabling a multi-tenant future for OCP. What happens in the PR:

1. All the entity tables that used a "name" as ID now use their integer ID as primary key, including in relation tables (bundles requiring sources, etc)
2. Entity names are now unique within a "tenant". Currently, only the "default" tenant exists, and it's not exposed through the HTTP API.
3. Almost all tables are recreated in the migration. This is done because it's tricky to update cross-table constraints in a portable way across sqlite/mysql/postgres.

> [!NOTE]
> Nothing changes for users of the CLI or HTTP API. Any differences to how OCP behaved before are considered bugs. 🐛 Please report them!

> [!WARNING]
> **Database Backups:**
> In MySQL, it seems impossible to run the migration in a transaction, since schema changes ("RENAME TABLE ...") _implicitly commit_ the transaction.
> **Please back up your database before running the migrations in this update.**

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -414,6 +414,7 @@ func sourcesDataGet[T any](ctx context.Context, d *Database, sourceName, path st
 
 		expr, err := authz.Partial(ctx, authz.Access{
 			Principal:  principal,
+			Tenant:     tenant,
 			Permission: "sources.data.read",
 			Resource:   "sources",
 			Name:       sourceName,
@@ -1720,7 +1721,7 @@ func (d *Database) UpsertToken(ctx context.Context, principal, tenant string, to
 			return err
 		}
 
-		return d.UpsertPrincipalTx(ctx, tx, Principal{Id: token.Name, Role: token.Scopes[0].Role})
+		return d.UpsertPrincipalTx(ctx, tx, Principal{Id: token.Name, Role: token.Scopes[0].Role, Tenant: tenant})
 	})
 }
 
--- a/internal/authz/authz.rego
+++ b/internal/authz/authz.rego
@@ -5,11 +5,13 @@ import rego.v1
 allow if {
 	data.principals.id == input.principal
 	data.principals.role == "administrator"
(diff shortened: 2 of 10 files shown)


.github/workflows/main.yaml
```diff
@@ -59,17 +61,26 @@ jobs:
           cache-dependency-path: go.sum
       - name: build
         run: make build
+      - name: build previous release
+        run: |
+          git worktree add ../prev v0.1.0
```
johanfylling: Is the previous version supposed to always be lagging one release behind, or is `v0.1.0` of special interest for a while now?

internal/database/database.go
```diff
@@ -1717,15 +1845,31 @@ func (d *Database) upsert(ctx context.Context, tx *sql.Tx, table string, columns
...
+			_, err := tx.ExecContext(ctx, query, values...)
+			if err != nil {
+				return 0, err
+			}
+			var id int
+			query = fmt.Sprintf("SELECT %[1]s.id FROM %[1]s JOIN tenants ON tenants.id = %[1]s.tenant_id AND %[1]s.%[2]s = %[3]s", table, primaryKey[0], d.arg(0))
+			return id, tx.QueryRowContext(ctx, query, values[0]).Scan(&id)
+		}
+		var id int
+		query += " RETURNING id"
```
johanfylling: TIL (rusty DB n00b :D)

e2e/cli/build_migrate_db.txtar
```diff
@@ -0,0 +1,111 @@
+[!env:E2E_OPACTL_PREV] stop 'E2E_OPACTL_PREV not set, not running migration test'
+
+# local example git repo
+cd repo
```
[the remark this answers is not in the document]
srenatus: That's created implicitly through the files specified below. txtar is a bit upside down that way 🙃

johanfylling (approved):

PR #211, opened 2025-12-01
Support POST method for http datasources

zsombix:
In Styra currently POST methods are supported when defining http datasources.

This change adds support for POST API calls simply by extending the `config` section of the datasource definition with the `method` and `body` fields.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,3 @@ opactl_*
 /files
 /.go
 /dist
-
-.env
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -14,3 +14,5 @@ opactl_*
 /files
 /.go
 /dist
+
+.env
```
srenatus: Can we keep that file as-is, please? There's nothing inside the repo that's using .env, I believe.

PR #212, opened 2025-12-02
database: support cockroachdb

srenatus:
This change brings support for CockroachDB as a database system. Use a dsn like this:

```yaml
database:
  sql:
    dsn: 'cockroachdb://crdbuser@192.168.120.2:33351/db?sslmode=disable'
    driver: cockroachdb
```

---------
All the tests pass after the migrations have been applied to a fresh CRDB instance and the migrations table looks as expected as well:
```
select * from schema_migrations;
 version | dirty
---------+-------
      19 | false
(1 row)
```

So I think we're good here. Remaining TODOs:
- [x] document decisions around migrations for future reference
- [x] check that all the constraints we set up are named, so we can refer to them in future migrations

### DIFF
--- a/internal/migrations/migrations_next.go
+++ b/internal/migrations/migrations_next.go
@@ -69,7 +69,7 @@ func crossTablesWithIDPKeys(offset int, dialect string) fs.FS {
 	return ocp_fs.MapFS(map[string]string{f: strings.Join(stmts, "; ")})
 }
 
-var v2Tables = []sqlTable{
+var v2Tables = []*sqlTable{
 	// tenants, new
 	createSQLTable("tenants").
 		WithIteration("ocp_v2").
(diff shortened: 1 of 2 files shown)


internal/database/database.go
```diff
@@ -302,9 +309,14 @@ func (d *Database) InitDB(ctx context.Context) error {
...
 		}
 
-	case d.config != nil && d.config.SQL != nil && (d.config.SQL.Driver == "postgres" || d.config.SQL.Driver == "pgx"):
+	case d.config.SQL.Driver == "postgres" || d.config.SQL.Driver == "pgx" || d.config.SQL.Driver == "cockroachdb":
 		dsn := os.ExpandEnv(d.config.SQL.DSN)
 		d.kind = postgres
+		if d.config.SQL.Driver == "cockroachdb" {
+			d.kind = cockroach // for tx specials
+			d.executeTx = crdb.ExecuteTx
+			dsn = "postgresql" + dsn[11:]
```
srenatus: eek that's bad. I'll fix that.

Issue #239, opened 2026-01-21
OCP builds bundle with a revision


bdjgs:
A bundle contains a `.manifest` file that has a revision field. The bundles build by OCP leaves this field empty instead of filling it with a unique value that can identify the version of the bundle.

Would be nice to that OCP builds a bundle with a unique revision value.

In this way, it is also possible for the OPA to expose via it's metrics which bundle it is currently running with.

**Extra:** optional
Currently, we are uploading bundles to an S3 storage, specifically MinIO. Currently, in MinIO, I can see the metadata field called X-Amz-Meta-Sha256, which I think is uploaded by OCP. I would be really nice if the solution to this task also uploads the revision as metadata to S3 (MinIO). 
The usecase is that it should be easy to find the specific bundle that the OPA is using in the S3 storage. So whatever solution that solves this is great 👍 I also noted that OPA actually logs the etag of the bundle upon activating it. However, that is not exposed in metrics. So maybe a better solution to the problem of linking bundle in S3 to the live bundle in OPA is that OPA exposes etag in metrics.

PR #240, opened 2026-01-21
set bundle revision to unix timestamp

bdjgs:
#239

### DIFF
--- a/internal/builder/builder.go
+++ b/internal/builder/builder.go
@@ -376,6 +376,7 @@ func (b *Builder) Build(ctx context.Context) error {
 		target = "plan"
 	}
 
+	fmt.Println("Building bundle with revision:", b.revision)
 	revision := ""
 	if b.revision == "time.now_ns()" {
 		revision = strconv.FormatInt(time.Now().UnixNano(), 10)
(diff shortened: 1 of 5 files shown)


srenatus:
I can fix the e2e tests, don't worry about those. I'd appreciate if you could appease the linter, though.

One question we have to figure out is if this -- as a default -- is going to be an acceptable change for everyone.

I wonder if we should make this configurable, and perhaps use Rego for that... basically

```yaml
bundles:
  my-bundle:
    revision: 'time.now_ns()'
    # ... other settings
```

What do you think? Since we do releases now, I'd be OK merging this, and turning it into :point_up: myself before the next release goes out.

bdjgs:
I also just realized I didnt make the PR a draft :) But thanks again for the quick feedback. 

My initial plan was also to make it configurable and I think your solution to include it in the bundle config is great 👍 

I will see if I can get it to work.

srenatus:
Sorry, no rush! Let me know when you get stuck 😃

bdjgs:
@srenatus I think this is a bit too much for me. As we are in the process of migrating from Styra to OCP, there keeps coming up other stuff internally. It would take too long time for me to complete this with my current understanding of the code base and db. 

There is no rush to complete this, it is not a critical functionality, atleast not for us. I will make sure that the issue includes all desired features and leave it for now. After migration is done, I might come back to this if it is still open.

As always, thanks for the quick feedback and all the help, it is really great 👍

srenatus:
I'll take over! Thanks for getting this started @bdjgs and thanks for your input @ghmer!

Issue #235, opened 2026-01-14
Pluggable Authorization Model for Data Layer


ashutosh-narkar:
Currently, the authorization logic within OCP is tightly coupled with the `database` package. The authorization policy is hardcoded via `//go:embed` in the `authz` package, making it difficult for consumers who use OCP as a library to override or bypass the authorization checks. For example, a parent application may have its own higher-level authorization system and need to treat the OCP data layer as a trusted subsystem.

#### Proposed Solution

This issue proposes refactoring the `authz` and `database` packages to support a pluggable authorization model. This will decouple the data layer from authorization logic, making the `database` package a more flexible and reusable library component.

1.  **Create an `Authorizer` Interface**: Introduce a new `authz.Authorizer` interface that defines the contract for all authorization modules.
    ```go
    type Authorizer interface {
        Check(ctx context.Context, tx *sql.Tx, fn ArgFn, access Access) bool
        Partial(ctx context.Context, access Access, extraColumnMappings map[string]columnRref) (Expr, error)
    }
    ```

2.  **Refactor `database.DB`**: Modify the `database.DB` struct to hold an `Authorizer`.
    ```go
    type DB struct {
        db         *sql.DB
        authorizer authz.Authorizer
    }
    ```

3.  **Introduce New Constructor**: Add a new constructor, `database.NewWithAuthorizer`, that allows consumers to inject their desired `Authorizer` implementation.

4.  **Provide Default `OPAAuthorizer`**: The existing OPA-based logic will be encapsulated in an `OPAAuthorizer` struct that implements the new interface.

We can ensure backwards compatibility by initializing the database instance to use the default `OPAAuthorizer`.

PR #246, opened 2026-01-28
Add Pluggable Authorization Model for Data Layer

ashutosh-narkar:
Currently, the authorization logic within OCP is tightly coupled with the `database` package. This change adds support for a pluggable authorization model in order to decouple the data layer from authorization logic.

Fixes: #235


pkg/authz/authz.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// AccessDescriptor describes the context of an access request.
+type AccessDescriptor interface {
+	GetPrincipal() string
+	GetTenant() string
+	GetResource() string
+	GetPermission() string
+	GetName() string
+}
```
srenatus: This gives off a faint protobuf smell, would it be much to ask to drop the `Get` and have your code provide a wrapping structure of some sort?

That kind of thing:
```go
// protoAccessAdapter adapts a ProtoAccessDescriptor to the AccessDescriptor interface.
type protoAccessAdapter struct {
	protoDesc ProtoAccessDescriptor
}

// NewAccessDescriptorFromProto creates an AccessDescriptor from a ProtoAccessDescriptor.
func NewAccessDescriptorFromProto(protoDesc ProtoAccessDescriptor) AccessDescriptor {
	return &protoAccessAdapter{
		protoDesc: protoDesc,
	}
}

// Implement the AccessDescriptor interface for protoAccessAdapter
func (a *protoAccessAdapter) Principal() string {
	return a.protoDesc.GetPrincipal()
}

func (a *protoAccessAdapter) Tenant() string {
	return a.protoDesc.GetTenant()
}

func (a *protoAccessAdapter) Resource() string {
	return a.protoDesc.GetResource()
}

func (a *protoAccessAdapter) Permission() string {
	return a.protoDesc.GetPermission()
}

func (a *protoAccessAdapter) Name() string {
	return a.protoDesc.GetName()
}
```

But I think for this to make sense, we'll have to change the access fields in the code _of this repo_ to use lower-case fields and uppercase accessors. 🤔

internal/authz/authz.go
```diff
@@ -211,9 +191,40 @@ type Access struct {
...
+	return a.Tenant
+}
+
+func (a Access) GetResource() string {
+	return a.Resource
+}
+
+func (a Access) GetPermission() string {
+	return a.Permission
+}
```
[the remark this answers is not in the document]
ashutosh-narkar: `ast.InterfaceToValue ` also needs exported fields IIRC.

pkg/authz/authz.go
ashutosh-narkar: Updated the interface now that the fields are lowercase.

```go
type AccessDescriptor interface {
	Principal() string
	Tenant() string
	Resource() string
	Permission() string
	Name() string
}
```

PR #262, opened 2026-02-05
Refactor credential resolution and add common sync interfaces

yi-chen-roger:
Consolidate duplicate credential resolution logic in gitsync and httpsync by introducing shared interfaces and combining legacy/non-legacy code paths.

Changes:
- Add pkg/sync package with Synchronizer and SecretProvider interfaces
- Add NewWithProvider constructors for external secret integration
- Refactor initClient() in httpsync to combine credential resolution paths
- Refactor auth() in gitsync to combine credential resolution paths
- Remove authFromMap() helper and related code duplication
- Add ClientSecret interface support in internal/config
- Simplify credential type conversion using shared authFromTyped()

This reduces code duplication and makes credential handling more maintainable by ensuring both legacy config-based and external provider-based credentials flow through the same conversion logic after initial resolution.

### DIFF
--- a/internal/config/secret.go
+++ b/internal/config/secret.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"cmp"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
@@ -472,10 +473,7 @@ type SecretAPIKey struct {
 }
 
 func (s *SecretAPIKey) Client(context.Context) (*http.Client, error) {
-	in := s.In
-	if in == "" {
-		in = "header" // default
-	}
+	in := cmp.Or(s.In, "header") // default
 
 	return wrappedClient(func(r *http.Request) *http.Request {
 		if in == "header" {
--- a/internal/gitsync/gitsync.go
+++ b/internal/gitsync/gitsync.go
@@ -96,17 +96,13 @@ func init() {
 //   - Rotate credentials without config changes
 //   - Audit secret access
 //   - Integrate with enterprise secret management systems
-//
-// SecretProvider is an alias to pkg/sync.SecretProvider for backward compatibility.
-// See pkg/sync package for interface documentation and supported credential types.
-type SecretProvider = pkgsync.SecretProvider
 
 type Synchronizer struct {
 	path           string
 	config         config.Git
 	gh             github
 	sourceName     string
-	secretProvider SecretProvider
+	secretProvider pkgsync.SecretProvider
 }
 
 // New creates a new Synchronizer instance. It is expected the threadpooling is outside of this package.
@@ -119,14 +115,7 @@ func New(path string, config config.Git, sourceName string) *Synchronizer {
 
 // WithSecretProvider configures the synchronizer to use an external SecretProvider for authentication.
 // This allows external projects to integrate their own secret management systems (e.g., Whisper, Vault).
-//
-// If provider is nil, credentials will be resolved from the config file.
-//
-// Example:
-//
-//	s := gitsync.New(path, config, sourceName).
-//	    WithSecretProvider(myProvider)
-func (s *Synchronizer) WithSecretProvider(provider SecretProvider) *Synchronizer {
+func (s *Synchronizer) WithSecretProvider(provider pkgsync.SecretProvider) *Synchronizer {
 	s.secretProvider = provider
 	return s
 }
@@ -275,14 +264,11 @@ func (s *Synchronizer) auth(ctx context.Context) (transport.AuthMethod, error) {
 
 	var typed any
 
-	// Resolve credentials to typed value
 	if s.secretProvider != nil {
-		// External SecretProvider integration
 		credMap, err := s.secretProvider.GetSecret(ctx, s.config.Credentials.Name)
 		if err != nil {
 			return nil, err
 		}
-		// Convert map to typed credential
 		secret := &config.Secret{
 			Name:  s.config.Credentials.Name,
 			Value: credMap,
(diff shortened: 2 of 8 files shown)


internal/config/secret.go
```diff
@@ -441,6 +448,48 @@ func (value *SecretTLSCert) ToConfig(ctx context.Context) (*tls.Config, error) {
...
+	Key   string `json:"key"`   // Header name or query param name (e.g., "X-API-Key")
+	Value string `json:"value"` // The API key value
+	In    string `json:"in"`    // Where to send the key: "header" or "query" (default: "header")
+}
+
+func (s *SecretAPIKey) Client(context.Context) (*http.Client, error) {
+	in := s.In
+	if in == "" {
+		in = "header" // default
+	}
```
srenatus: [nit]
```suggestion
	in := cmp.Or(s.In, "header") // default
```

internal/httpsync/httpsync.go
```diff
@@ -14,34 +14,57 @@ import (
...
+	client         *http.Client
+	s3Client       *s3.Client
 }
 
 type HeaderSetter interface {
 	SetHeader(*http.Request) error
 }
 
-func New(path, url, method, body string, headers map[string]any, credentials *config.SecretRef) *HttpDataSynchronizer {
+// SecretProvider is an alias to pkg/sync.SecretProvider for backward compatibility.
```
srenatus: 💭 Backward compatibility with what? Since this is an internal package, why do bother about the `type SecretProvider` at all? 🤔

internal/httpsync/httpsync.go
```diff
@@ -14,34 +14,57 @@ import (
...
+// See pkg/sync package for interface documentation and supported credential types.
+type SecretProvider = pkgsync.SecretProvider
+
+func New(path string, url string, method string, body string, headers map[string]any, credentials *config.SecretRef) *HttpDataSynchronizer {
```
srenatus: [nit] Please revert to what we had before,
```suggestion
func New(path, url, method, body string, headers map[string]any, credentials *config.SecretRef) *HttpDataSynchronizer {
```

internal/httpsync/httpsync.go
```diff
@@ -14,34 +14,57 @@ import (
...
+// This allows external projects to integrate their own secret management systems (e.g., Whisper, Vault).
+//
+// If credentialName is empty, the provider configuration is ignored and the synchronizer
+// will use unauthenticated requests (or fall back to credentials configured via New()).
+//
+// Example:
+//
+//	s := httpsync.New(path, url, method, body, headers, nil).
+//	    WithSecretProvider("my-secret", myProvider)
+func (s *HttpDataSynchronizer) WithSecretProvider(credentialName string, provider SecretProvider) *HttpDataSynchronizer {
```
srenatus: The code example and the arguments make me think that the _name_ was for the _provider_. Perhaps we can make it more explicit, like
```go
s := httpsync.New(path, url, method, body, headers, nil).WithSecretFromProvider("my-secret", myProvider)
```

or 
```go
s := httpsync.New(path, url, method, body, headers, nil).WithProvider(myProvider).WithSecret("my-secret")
```

Alternatively, we model this after `gitsync`, which passes the credentialName differently, it seems.

yi-chen-roger: update the signature to 
```
func (s *HttpDataSynchronizer) WithSecretProvider(provider SecretProvider) *HttpDataSynchronizer 
```
so, the usage will be 

```
syncer := httpsync.New(path, url, method, body, headers, credentials)
if provider != nil {
syncer = syncer.WithSecretProvider(provider)
}
```
I think now WithSecretProvider name makes more sense, and now the name and signature is consistent with gitsync

internal/gitsync/gitsync.go
```diff
@@ -117,10 +117,18 @@ func New(path string, config config.Git, sourceName string) *Synchronizer {
...
-	return &Synchronizer{path: path, config: config, sourceName: sourceName, secretProvider: provider}
+// WithSecretProvider configures the synchronizer to use an external SecretProvider for authentication.
+// This allows external projects to integrate their own secret management systems (e.g., Whisper, Vault).
+//
+// If provider is nil, credentials will be resolved from the config file.
+//
+// Example:
+//
+//	s := gitsync.New(path, config, sourceName).
+//	    WithSecretProvider(myProvider)
```
srenatus: Let's remove this, it adds no value

pkg/gitsync/secretprovider.go
```diff
@@ -1,7 +1,7 @@
...
 
-import internalgitsync "github.com/open-policy-agent/opa-control-plane/internal/gitsync"
+import pkgsync "github.com/open-policy-agent/opa-control-plane/pkg/sync"
 
-// SecretProvider is re-exported from internal/gitsync for external use.
-// See internal/gitsync package for interface documentation and supported credential types.
-type SecretProvider = internalgitsync.SecretProvider
+// SecretProvider is re-exported from pkg/sync for external use.
+// See pkg/sync package for interface documentation and supported credential types.
+type SecretProvider = pkgsync.SecretProvider
```
srenatus: Now it's getting weird: why do we export this a second time?

CLAUDE.md
srenatus: Please remove this file

internal/httpsync/httpsync.go
```diff
@@ -34,6 +36,10 @@ type HeaderSetter interface {
 	SetHeader(*http.Request) error
 }
 
+// SecretProvider is an alias to pkg/sync.SecretProvider.
+// See pkg/sync package for interface documentation and supported credential types.
+type SecretProvider = pkgsync.SecretProvider
```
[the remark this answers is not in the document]
yi-chen-roger: you are right, I will remove it

yi-chen-roger:
No problem at all! I really appreciate the thorough feedback. 

Updated the code base on your comment @srenatus

srenatus (approved):
LGTM thanks!

PR #314, opened 2026-04-01
builder: record placeholder roots for sources with `prefix`

srenatus:
...while respecting `path` and transitive application thereof.

Also adjusts fs sources without actual policies in them (see added tests).

Fixes #312.

### DIFF
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"os"
 	"slices"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/builder/builder_test.go
```diff
@@ -631,6 +633,97 @@ func TestBuilder(t *testing.T) {
...
+				{
+					name:  "leaf",
+					files: map[string]string{},
+				},
+			},
+			expRoots: []string{"top/inner"},
+		},
+		{
+			note: "empty source with path and prefix that cancel out",
+			sources: []sourceMock{
```
mefarazath, line 709: Is this a case where despite source "middle" having a prefix "x", the final manifest will not have any roots since "primary" source selects data from "data.x" and doesn't define a prefix?

srenatus: It's about removing what was added down the transitive path. I've tried clarifying the test.

mefarazath: I understand the test case better now :)

PR #315, opened 2026-04-01
Add bundle lifecycle status tracking (build, push)

ashutosh-narkar:
This change adds support for tracking the status of a bundle
through the build and push phases. The sync phase
is not tracked as we need a bundle revision to keep
track of a tenant+bundle+revision combo.

### DIFF
--- a/internal/authz/authz_test.rego
+++ b/internal/authz/authz_test.rego
@@ -19,6 +19,7 @@ read_permissions := {
 	"secrets.view",
 	"tokens.view",
 	"sources.data.read",
+	"bundles.statuses.view",
 }
 
 test_viewer_can_view_anything if {
--- a/internal/database/bundle_status.go
+++ b/internal/database/bundle_status.go
@@ -1,0 +1,197 @@
+package database
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/open-policy-agent/opa-control-plane/pkg/config"
+)
+
+const MaxBundleStatusRetention = 10
+
+// UpsertBundleStatus creates or updates a bundle status record with the given phase and status.
+// For a given tenant+bundle+revision combination, only one record exists.
+// If a record already exists for the combination, it updates the phase and status.
+// Old records beyond the retention limit (MaxBundleStatusRetention) are cleaned up in the same transaction.
+func (d *Database) UpsertBundleStatus(ctx context.Context, tenant, bundle, revision, phase, status, errMsg string) (int, error) {
+	if revision == "" {
+		return 0, errors.New("error inserting bundle status: revision is required")
+	}
+
+	var id int
+	err := tx1(ctx, d, func(tx *sql.Tx) error {
+
+		bundleID, err := d.lookupID(ctx, tx, tenant, "bundles", bundle)
+		if err != nil {
+			return fmt.Errorf("error looking up bundle %s: %w", bundle, err)
+		}
+
+		resID, err := d.upsertReturning(ctx, true, true, tx, tenant, "bundles_statuses", []string{"bundle_id", "revision", "phase", "status", "error_message"}, []string{"bundle_id", "revision"},
+			bundleID, revision, phase, status, errMsg)
+		if err != nil {
+			return fmt.Errorf("error inserting bundle status: %w", err)
+		}
+		id = resID
+
+		// Cleanup: keep only the last MaxBundleStatusRetention records per tenant+bundle_id.
+		// This runs in the same transaction to ensure consistency.
+		//
+		// Find the ID of the oldest record we want to keep and then delete everything older.
+		var cutoffID int64
+		err = tx.QueryRowContext(ctx,
+			fmt.Sprintf(
+				`SELECT bundles_statuses.id FROM bundles_statuses
+				 JOIN tenants ON tenants.id = bundles_statuses.tenant_id
+				 WHERE tenants.name = %s AND bundles_statuses.bundle_id = %s
+				 ORDER BY bundles_statuses.id DESC
+				 LIMIT 1 OFFSET %s`,
+				d.arg(0), d.arg(1), d.arg(2),
+			),
+			tenant, bundleID, MaxBundleStatusRetention-1,
+		).Scan(&cutoffID)
+
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf("error finding cutoff for bundle status cleanup: %w", err)
+		}
+
+		// If we found a cutoff, delete all records older than it.
+		if err == nil {
+			_, err = tx.ExecContext(ctx,
+				fmt.Sprintf(
+					`DELETE FROM bundles_statuses WHERE bundle_id = %s AND id < %s`,
+					d.arg(0), d.arg(1),
+				),
+				bundleID, cutoffID,
+			)
+			if err != nil {
+				return fmt.Errorf("error cleaning up old bundle statuses: %w", err)
+			}
+		}
+
+		return nil
+	})
+	return id, err
+}
+
+// GetBundleStatus retrieves a single bundle state record by ID.
+func (d *Database) GetBundleStatus(ctx context.Context, id int) (*config.BundleStatus, error) {
+	var s config.BundleStatus
+	err := tx1(ctx, d, func(tx *sql.Tx) error {
+		err := tx.QueryRowContext(ctx,
+			`SELECT id, tenant_id, bundle_id, revision, phase, status, error_message, created_at
+			 FROM bundles_statuses
+			 WHERE id = `+d.arg(0),
+			id,
+		).Scan(
+			&s.ID, &s.TenantID, &s.BundleID, &s.Revision, &s.Phase, &s.Status,
+			&s.ErrorMessage, &s.CreatedAt,
+		)
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				return ErrNotFound
+			}
+			return fmt.Errorf("error querying bundle status by id: %w", err)
+		}
+		return nil
+
+	})
+
+	return &s, err
+}
+
+// GetLatestBundleStatus returns the most recent status record for the given tenant and bundle
+// across all revisions. Returns nil and ErrNotFound if no records exist.
+func (d *Database) GetLatestBundleStatus(ctx context.Context, principal, tenant, bundle string) (*config.BundleStatus, error) {
+	statuses, err := d.ListBundleStatuses(ctx, principal, tenant, bundle, "", 1)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(statuses) == 0 {
+		return nil, ErrNotFound
+	}
+
+	return statuses[0], nil
+}
+
+// ListBundleStatuses returns bundle status records for the given tenant and bundle,
+// ordered by bundle status id DESC. If revision is provided, results are filtered by revision.
+// If revision is empty, returns records across all revisions.
+// A limit of 0 defaults to MaxBundleStatusRetention; values above MaxBundleStatusRetention are capped.
+func (d *Database) ListBundleStatuses(ctx context.Context, principal, tenant, bundle, revision string, limit int) ([]*config.BundleStatus, error) {
+
+	// Apply default and cap for limit.
+	if limit <= 0 {
+		limit = MaxBundleStatusRetention
+	}
+	if limit > MaxBundleStatusRetention {
+		limit = MaxBundleStatusRetention
+	}
+
+	var statuses []*config.BundleStatus
+
+	err := tx1(ctx, d, func(tx *sql.Tx) error {
+		ad := d.accessFactory().WithPrincipal(principal).WithTenant(tenant).WithResource("bundles_statuses").WithPermission("bundles.statuses.view").WithName(bundle)
+		if !d.authorizer.Check(ctx, tx, d.arg, ad) {
+			return ErrNotAuthorized
+		}
+
+		bundleID, err := d.lookupID(ctx, tx, tenant, "bundles", bundle)
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf("error looking up bundle %s: %w", bundle, err)
+		}
+
+		var query string
+		var args []any
+
+		if revision != "" {
+			query = fmt.Sprintf(
+				`SELECT bundles_statuses.id, tenant_id, bundle_id, revision, phase, status, error_message, created_at
+				 FROM bundles_statuses
+				 JOIN tenants ON tenants.id = bundles_statuses.tenant_id
+				 WHERE tenants.name = %s AND bundle_id = %s AND revision = %s
+				 ORDER BY bundles_statuses.id DESC
+				 LIMIT %s`,
+				d.arg(0), d.arg(1), d.arg(2), d.arg(3),
+			)
+			args = []any{tenant, bundleID, revision, limit}
+		} else {
+			query = fmt.Sprintf(
+				`SELECT bundles_statuses.id, tenant_id, bundle_id, revision, phase, status, error_message, created_at
+				 FROM bundles_statuses
+				 JOIN tenants ON tenants.id = bundles_statuses.tenant_id
+				 WHERE tenants.name = %s AND bundle_id = %s
+				 ORDER BY bundles_statuses.id DESC
+				 LIMIT %s`,
+				d.arg(0), d.arg(1), d.arg(2),
+			)
+			args = []any{tenant, bundleID, limit}
+		}
+
+		rows, err := tx.Query(query, args...)
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var s config.BundleStatus
+			if err := rows.Scan(
+				&s.ID, &s.TenantID, &s.BundleID, &s.Revision, &s.Phase, &s.Status,
+				&s.ErrorMessage, &s.CreatedAt,
+			); err != nil {
+				return fmt.Errorf("error scanning bundle status: %w", err)
+			}
+			s.BundleName = bundle
+			statuses = append(statuses, &s)
+		}
+		if err := rows.Err(); err != nil {
+			return fmt.Errorf("error iterating bundle statuses: %w", err)
+		}
+		return nil
+	})
+
+	return statuses, err
+}
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -291,6 +291,7 @@ func (d *Database) InitDB(ctx context.Context) error {
 					return err
 				}
 				cfg.MultiStatements = true // required for migrations to work
+				cfg.ParseTime = true       // required for scanning time.Time columns
 			} else {
 				cfg = &mysqldriver.Config{
 					User:                    dbUser,
@@ -302,6 +303,7 @@ func (d *Database) InitDB(ctx context.Context) error {
 					AllowOldPasswords:       true,
 					TLSConfig:               tlsConfigName,
 					MultiStatements:         true, // required for migrations
+					ParseTime:               true, // required for scanning time.Time columns
 				}
 
 				_ = cfg.Apply(mysqldriver.BeforeConnect(func(ctx context.Context, config *mysqldriver.Config) error {
@@ -369,6 +371,7 @@ func (d *Database) InitDB(ctx context.Context) error {
 			return err
 		}
 		cfg.MultiStatements = true // required for migrations
+		cfg.ParseTime = true       // required for scanning time.Time columns
 
 		cfg.TLS, err = tlsConfig(ctx, d.config.SQL.Credentials, cfg.TLS)
 		if err != nil {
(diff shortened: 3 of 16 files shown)


pkg/service/service_test.go
```diff
@@ -127,6 +129,204 @@ func TestRequirementsWithOverrides(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	stopped := make(chan struct{})
+	g.Go(func() error {
+		defer close(stopped)
+		return svc.Run(ctx)
+	})
+
+	time.Sleep(time.Millisecond * 500)
```
srenatus, line 307: Can you please use `testing/synctest`? It should let us run the same test without an _actual_ sleep. (Same as other test.)

pkg/service/service_test.go
```diff
@@ -127,6 +129,204 @@ func TestRequirementsWithOverrides(t *testing.T) {
...
+	}
+
+	svc := service.New().
+		WithConfig(cfg).
+		WithPersistenceDir(filepath.Join(tempDir, "data")).
+		WithMigrateDB(true).
+		WithLogger(log)
+
+	var g errgroup.Group
+	ctx, cancel := context.WithCancel(context.Background())
```
srenatus: [nit] Let's be consistent wth `ctx := t.Context()`. Also we can then use this below with `GetLatestBundleStatus`.

pkg/service/worker.go
```diff
@@ -94,6 +98,21 @@ func (worker *BundleWorker) WithInterval(d config.Duration) *BundleWorker {
 	return worker
 }
 
+func (worker *BundleWorker) WithDatabase(database *database.Database) *BundleWorker {
+	worker.database = database
+	return worker
+}
+
+func (worker *BundleWorker) WithPrincipal(principal string) *BundleWorker {
```
srenatus: Why does the bundle worker need a principal..? 🤔 I think it would make sense if `UpsertBundleStatus` didn't require authz. It's not exposed through any http handler.

internal/database/database.go
```diff
@@ -2022,6 +2025,73 @@ func (d *Database) upsertReturning(ctx context.Context, returning bool, tx *sql.
...
 }
 
+// upsertReturningLastInsertID is similar to upsertReturning but uses MySQL's LAST_INSERT_ID()
+// to retrieve the inserted/updated row's ID instead of performing a follow-up SELECT query.
+// For MySQL, it adds `id = LAST_INSERT_ID(id)` to the ON DUPLICATE KEY UPDATE clause so that
+// LastInsertId() returns the correct ID for both inserts and updates.
+// For sqlite, postgres, and cockroach, it behaves identically to upsertReturning with returning=true.
+func (d *Database) upsertReturningLastInsertID(ctx context.Context, tx *sql.Tx, tenant, table string, columns []string, primaryKey []string, values ...any) (int64, error) {
+	valueArgs := d.args(len(columns))
+	if tenant != "" { // not a relation-only table
```
srenatus: This is taken from the generic upsert helper, I think? We don't need to deal with it like that here, I suppose.

internal/database/database.go
```diff
@@ -2022,6 +2025,73 @@ func (d *Database) upsertReturning(ctx context.Context, returning bool, tx *sql.
 	}
 }
 
+// upsertReturningLastInsertID is similar to upsertReturning but uses MySQL's LAST_INSERT_ID()
+// to retrieve the inserted/updated row's ID instead of performing a follow-up SELECT query.
+// For MySQL, it adds `id = LAST_INSERT_ID(id)` to the ON DUPLICATE KEY UPDATE clause so that
+// LastInsertId() returns the correct ID for both inserts and updates.
```
srenatus: Would it perhaps make sense to adjust `upsertReturning`? Sounds like this twist would be applicable across the tables.

srenatus (commented):
Couple of early comments! 👍

pkg/service/service_test.go
ashutosh-narkar: I'd appreciate some help. I'm getting the error `panic: deadlock: main bubble goroutine has exited but blocked goroutines remain [recovered, repanicked]` when I use `synctest.Test` and `synctest.Wait`.  I'm probably doing something wrong here.

pkg/service/service_test.go
ashutosh-narkar: Updated

pkg/service/worker.go
ashutosh-narkar: Makes sense. Removed authz for `UpsertBundleStatus`

internal/database/database.go
ashutosh-narkar: removed `upsertReturningLastInsertID` and rolled this into `upsertReturning `

internal/database/database.go
ashutosh-narkar: removed `upsertReturningLastInsertID` and rolled this into `upsertReturning`

pkg/service/service_test.go
srenatus: Ah, let's leave it as-is then. We can clean this up later.

internal/authz/authz_test.rego
```diff
@@ -19,6 +19,7 @@ read_permissions := {
 	"secrets.view",
 	"tokens.view",
 	"sources.data.read",
+	"bundles_statuses.view",
```
srenatus: Given that we have `sources.data.read`, wouldn't `bundles.status.view` fit in slightly better?

internal/database/database.go
```diff
@@ -1945,20 +1948,20 @@ func (d *Database) lookupID(ctx context.Context, tx *sql.Tx, tenant, table, name
...
 }
 
 func (d *Database) upsertRel(ctx context.Context, tx *sql.Tx, table string, columns []string, primaryKey []string, values ...any) error {
-	_, err := d.upsertReturning(ctx, false, tx, "", table, columns, primaryKey, values...)
+	_, err := d.upsertReturning(ctx, false, false, tx, "", table, columns, primaryKey, values...)
 	return err
 }
 
-func (d *Database) upsertReturning(ctx context.Context, returning bool, tx *sql.Tx, tenant, table string, columns []string, primaryKey []string, values ...any) (int, error) {
+func (d *Database) upsertReturning(ctx context.Context, returning, returningLastInsertID bool, tx *sql.Tx, tenant, table string, columns []string, primaryKey []string, values ...any) (int, error) {
```
srenatus, line 1964: Hm! I'm surprised there's a need to differentiate "returning" and "returningLastInsertID". I took the former to mean the same when it was introduced. When do we only want one of them?

pkg/builder/builder.go
```diff
@@ -214,6 +214,10 @@ func (b *Builder) WithRevisionFunc(fn func(fs.FS) (string, error)) *Builder {
 	return b
 }
 
+func (b *Builder) Revision() string {
```
srenatus, line 217: Is this always safe to call? I think `b.revision` might only get populated at bundle build time...? 🤔

internal/server/server_test.go
```diff
@@ -1167,3 +1169,172 @@ func (tr *testResponse) ExpectBody(x any) *testResponse {
...
+			t.Run("empty result for unknown bundle", func(t *testing.T) {
+				var resp types.BundleStatusListResponseV1
+				ts.Request("GET", "/v1/bundles/nonexistent/status", "", ownerKey).
+					ExpectStatus(http.StatusOK).
+					ExpectBody(&resp)
+
+				assert.Empty(t, resp.Result)
+			})
+
+			t.Run("unauthorized", func(t *testing.T) {
```
srenatus, line 1336: Should we check some cross-tenant request, perhaps?

srenatus (commented):
Had a quick look, it looks good, some nitpicks only!

pkg/builder/builder.go
ashutosh-narkar: We're only tracking build and upload state as we need the revision to be non-empty.

ashutosh-narkar: We can't track sync state due to this.

internal/database/database.go
ashutosh-narkar: IIUC the special handling is only needed for MySQL. In our case, we want the ID for both inserts and updates. Since the other databases support the `RETURNING` clause, they return the correct ID. MySQL will return 0 or something for the update case. Hence this is a workaround to get the correct ID on update.

internal/server/server_test.go
ashutosh-narkar: Added a test case for this.

internal/authz/authz_test.rego
ashutosh-narkar: Updated to `bundles.status.view`

internal/server/server.go
```diff
@@ -66,6 +66,8 @@ func (s *Server) Init() *Server {
 	setup("GET", "/v1/bundles/{bundle}", s.v1BundlesGet)
 	setup("PUT", "/v1/bundles/{bundle}", s.v1BundlesPut)
 	setup("DELETE", "/v1/bundles/{bundle}", s.v1BundlesDelete)
+	setup("GET", "/v1/bundles/{bundle}/status/latest", s.v1BundleStatusLatestGet)
+	setup("GET", "/v1/bundles/{bundle}/status", s.v1BundleStatusList)
```
johanfylling, line 70: Will there be a follow up [docs](https://www.openpolicyagent.org/docs/ocp/api-reference#bundle-management) update for these new endpoints, or is that still too early?

internal/server/server_test.go
```diff
@@ -1167,3 +1169,199 @@ func (tr *testResponse) ExpectBody(x any) *testResponse {
...
+			db := initTestDB(t, databaseConfig.Database(t, ctr).Database)
+			ts := initTestServer(t, db)
+			defer ts.Close()
+
+			if err := db.UpsertPrincipal(ctx, principal); err != nil {
+				t.Fatal(err)
+			}
+
+			const ownerKey = "test-owner-key"
+			if err := db.UpsertToken(ctx, "internal", "default", &config.Token{
```
johanfylling, line 1196: [nit] to have less "magic" strings to track down

```suggestion
			if err := db.UpsertToken(ctx, principal.Id, principal.Tenant, &config.Token{
```

Same in other places.

internal/server/server_test.go
```diff
@@ -1167,3 +1169,199 @@ func (tr *testResponse) ExpectBody(x any) *testResponse {
...
+				t.Fatal(err)
+			}
+
+			if err := db.UpsertBundle(ctx, "internal", "default", &config.Bundle{Name: "bundle1"}); err != nil {
+				t.Fatal(err)
+			}
+
+			if _, err := db.UpsertBundleStatus(ctx, "default", "bundle1", "rev1", "build", "in_progress", ""); err != nil {
+				t.Fatal(err)
+			}
```
johanfylling, line 1210: Are there missing assertions following this status update, or this is here to show that we only get the immediately following status update when querying? Maybe there is value in asserting both statuses(?)

internal/server/server.go
johanfylling, line 70: Given that we're returning a list of statuses I suppose this could also be `.../statuses` if we want to be sticklers for naming consistency; but I'm guessing we're seeing this as a general "status" endpoint for the bundle, and the actual return type is irrelevant(?)
(also, `status` kinda reads better 😄)

internal/server/server.go
```diff
@@ -230,6 +232,61 @@ func (s *Server) v1BundlesDelete(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	principal, tenant := s.auth(r)
+	status, err := s.db.GetLatestBundleStatus(ctx, principal, tenant, name)
+	if err != nil {
+		errorAuto(w, err)
+		return
+	}
+
+	resp := types.BundleStatusGetResponseV1{Result: status}
```
johanfylling, line 251: If there are no status records, we're gonna return an empty dictionary (`{}`)? Maybe we should add a test for this to not have future regressions on this behavior.

internal/server/server.go
```diff
@@ -230,6 +232,61 @@ func (s *Server) v1BundlesDelete(w http.ResponseWriter, r *http.Request) {
...
+	JSONOK(w, resp, pretty(r))
+}
+
+func (s *Server) v1BundleStatusList(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	name, err := url.PathUnescape(r.PathValue("bundle"))
+	if err != nil {
+		ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, err)
```
johanfylling, line 260: A general question I don't think is specific to this PR: looks like we tend to return internal errors to the client-side; maybe we shouldn't do that, and instead commit those to log and instead return a lookup identifier 🤔?

@srenatus, is the security considerations here that the client is trusted? (I suppose authn/authz will always have preceded returning any error details, but does bundle access also automatically mean you have access to more general info like this?)

internal/database/bundle_status.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		return nil
+
+	})
+
+	return &s, err
+}
+
+// GetLatestBundleStatus returns the most recent status record for the given tenant and bundle
+// across all revisions. Returns nil if no records exist.
```
johanfylling: Looks like we're also returning an error on no records, which is probably the more salient parameter to highlight.

internal/server/server.go
johanfylling: Scratch that, I was misled by the func doc 😄. But we should still assert this behavior in testing.

internal/database/bundle_status.go
```diff
@@ -0,0 +1,197 @@
...
+func (d *Database) GetLatestBundleStatus(ctx context.Context, principal, tenant, bundle string) (*config.BundleStatus, error) {
+	statuses, err := d.ListBundleStatuses(ctx, principal, tenant, bundle, "", 1)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(statuses) == 0 {
+		return nil, ErrNotFound
```
johanfylling, line 113: Would there be a point in differentiating between a missing bundle and no status records by actually returning an empty value here? 
But maybe the client-side doesn't care about that distinction; and if it did, it could also HEAD the bundle endpoint directly to figure it out.

internal/database/bundle_status.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	return statuses[0], nil
+}
+
+// ListBundleStatuses returns bundle status records for the given tenant and bundle,
+// ordered by created_at DESC. If revision is provided, results are filtered by revision.
```
johanfylling: `created_at` is when individual status records were created?

internal/server/server_test.go
```diff
@@ -1167,3 +1169,199 @@ func (tr *testResponse) ExpectBody(x any) *testResponse {
...
+
+			t.Run("success", func(t *testing.T) {
+				var resp types.BundleStatusListResponseV1
+				ts.Request("GET", "/v1/bundles/bundle1/status", "", ownerKey).
+					ExpectStatus(http.StatusOK).
+					ExpectBody(&resp)
+
+				assert.Len(t, resp.Result, 2) // For a given tenant+bundle+revision combination, only one record exists.
```
johanfylling, line 1292: Should we not also assert record content?

internal/server/server_test.go
```diff
@@ -1167,3 +1169,199 @@ func (tr *testResponse) ExpectBody(x any) *testResponse {
...
+			const otherOwnerKey = "other-test-owner-key"
+			if err := db.UpsertToken(ctx, otherPrincipal, otherTenant, &config.Token{
+				Name:   "othertestowner",
+				APIKey: otherOwnerKey,
+				Scopes: []config.Scope{{Role: "viewer"}},
+			}); err != nil {
+				t.Fatal(err)
+			}
+
+			t.Run("other tenant unauthorized", func(t *testing.T) {
```
johanfylling, line 1362: To really cross our t:s, maybe we could additionally add overlapping bundle + revision to the other tenant to assert separation? (or have I misunderstood how tenants are partitioned?)

internal/database/bundle_status.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+		}
+
+		return nil
+	})
+	return id, err
+}
+
+// GetBundleStatus retrieves a single bundle state record by ID.
+func (d *Database) GetBundleStatus(ctx context.Context, id int) (*config.BundleStatus, error) {
```
johanfylling, line 79: Looks like this is only used in tests. This is for when using OCP as a lib?

internal/database/bundle_status.go
```diff
@@ -0,0 +1,197 @@
...
+				 WHERE tenants.name = %s AND bundles_statuses.bundle_id = %s
+				 ORDER BY bundles_statuses.id DESC
+				 LIMIT 1 OFFSET %s`,
+				d.arg(0), d.arg(1), d.arg(2),
+			),
+			tenant, bundleID, MaxBundleStatusRetention-1,
+		).Scan(&cutoffID)
+
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf("error finding cutoff for bundle status cleanup: %w", err)
```
johanfylling, line 56: But we've successfully upserted the status, right? In the long run, this would possibly cause an ever-growing DB which could have additional performance implications; but is this cause to fail the ongoing operation? To me, this looks more like something to log a warning for; or to provide a mechanism for the call-site to determine error severity.

internal/database/bundle_status.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		// If we found a cutoff, delete all records older than it.
+		if err == nil {
+			_, err = tx.ExecContext(ctx,
+				fmt.Sprintf(
+					`DELETE FROM bundles_statuses WHERE bundle_id = %s AND id < %s`,
```
johanfylling, line 63: I forget, are bundle ID:s unique server-wide, across tenants? If not, might we be deleting records for another tenant with here?

pkg/service/worker.go
```diff
@@ -203,9 +216,23 @@ func (w *BundleWorker) Execute(ctx context.Context) time.Time {
 
 	if err := b.Build(ctx); err != nil {
 		w.log.Warnf("failed to build a bundle %q: %v", w.bundleConfig.Name, err)
+
+		if b.Revision() != "" {
+			_, err := w.database.UpsertBundleStatus(ctx, w.tenant, w.bundleConfig.Name, b.Revision(), BuildPhaseBuild.String(), BuildStateBuildFailed.String(), err.Error())
+			if err != nil {
+				w.log.Warnf("failed to track bundle build state %q: %v", w.bundleConfig.Name, err)
```
johanfylling, line 223: Ok, so we don't abort on status error 👍.

pkg/service/worker.go
```diff
@@ -214,10 +241,24 @@ func (w *BundleWorker) Execute(ctx context.Context) time.Time {
 				return w.report(ctx, BuildStateSuccess, startTime, nil)
 			}
 			w.log.Warnf("failed to upload bundle %q: %v", w.bundleConfig.Name, err)
+
+			if b.Revision() != "" {
+				_, err := w.database.UpsertBundleStatus(ctx, w.tenant, w.bundleConfig.Name, b.Revision(), BuildPhasePush.String(), BuildStatePushFailed.String(), err.Error())
```
johanfylling, line 246: We only need to report on the ultimate success/failure of each phase, not that they're in progress?

pkg/service/worker.go
```diff
@@ -203,9 +216,23 @@ func (w *BundleWorker) Execute(ctx context.Context) time.Time {
...
 	}
 
+	if b.Revision() != "" {
+		_, err := w.database.UpsertBundleStatus(ctx, w.tenant, w.bundleConfig.Name, b.Revision(), BuildPhaseBuild.String(), BuildStateSuccess.String(), "")
+		if err != nil {
+			w.log.Warnf("failed to track bundle build state %q: %v", w.bundleConfig.Name, err)
+		}
+	}
+
 	if w.storage != nil {
```
johanfylling, line 236: It's up to the client to know that storage has been configured, and therefore when to expect a `PUSH` phase?
Would there be value in tracking each phase separately, or that would just be unwarranted complexity?

pkg/service/service_test.go
```diff
@@ -127,6 +129,204 @@ func TestRequirementsWithOverrides(t *testing.T) {
...
+
+	if status.Phase != service.BuildPhasePush.String() {
+		t.Fatalf("expected bundle phase %v but got %v", service.BuildPhasePush.String(), status.Phase)
+	}
+
+	if status.Status != service.BuildStatePushFailed.String() {
+		t.Fatalf("expected bundle status %v but got %v", service.BuildStatePushFailed.String(), status.Status)
+	}
+}
+
```
johanfylling, line 329: Should we also add a test case for when there is no storage, and we only expect `BUILD`-phase status?

johanfylling (approved):
LGTM 👍

internal/server/server.go
ashutosh-narkar: Once we merge this, I will create a PR to update the docs.

ashutosh-narkar: > also, `status` kinda reads better

👍

internal/server/server_test.go
ashutosh-narkar: Updated.

internal/server/server_test.go
ashutosh-narkar: We are testing that we get the last inserted status for a tenant+bundle combo. So we expect to get only one status which we assert later in the test.

internal/server/server.go
ashutosh-narkar: This behavior is consistent with `v1BundlesGet`. We have a test for this in internal/database/bundle_status_test.go (see `TestGetLatestBundleStatus`)

internal/database/bundle_status.go
ashutosh-narkar: Updated the description.

internal/database/bundle_status.go
ashutosh-narkar: I've tried to keep this aligned with existing behavior eg.`v1BundlesGet`

internal/database/bundle_status.go
ashutosh-narkar: I've updated this comment. We are returning records ordered by auto-incrementing bundle status id to get consistent behavior across all database dialects.

internal/server/server_test.go
ashutosh-narkar: Added record assertion.

internal/server/server_test.go
ashutosh-narkar: We can't test this atm due to https://github.com/open-policy-agent/opa-control-plane/blob/main/internal/server/server.go#L711

internal/database/bundle_status.go
ashutosh-narkar: Correct. Could come in handy.

internal/database/bundle_status.go
ashutosh-narkar: Fair. Since the clean up is happening in the same transaction as the insert, it will get aborted on error. We can revisit this if this becomes an issue.

internal/database/bundle_status.go
ashutosh-narkar: I think so. This is the same pattern used for deleting other resources.

pkg/service/worker.go
ashutosh-narkar: Only success/failure atm

pkg/service/worker.go
ashutosh-narkar: >It's up to the client to know that storage has been configured, and therefore when to expect a PUSH phase?

Yes. So clients can plugin their own storage implementation. 

> Would there be value in tracking each phase separately, or that would just be unwarranted complexity?

We currently track `build` and `push` phases. One phase for a tenant+bundle+revision combo.

pkg/service/service_test.go
ashutosh-narkar: The service will create a storage if the user does not provide one.

PR #316, opened 2026-04-01
pkg/service: add WithAuthorizer and expose logging package

yi-chen-roger:
Add WithAuthorizer to Service so external consumers can override the default OPAuthorizer (e.g. with a bypass authorizer). The authorizer is persisted across initDB since migrations.Run() creates a new database instance that resets to the default authorizer.

Also expose the internal logging package via pkg/logging so external consumers can create loggers for WithLogger without importing internal packages.

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"log/slog"
 	"maps"
 	"path"
 	"path/filepath"
@@ -32,6 +33,7 @@ import (
 	"github.com/open-policy-agent/opa-control-plane/internal/progress"
 	"github.com/open-policy-agent/opa-control-plane/internal/s3"
 	"github.com/open-policy-agent/opa-control-plane/internal/sqlsync"
+	ext_authz "github.com/open-policy-agent/opa-control-plane/pkg/authz"
 	"github.com/open-policy-agent/opa-control-plane/pkg/builder"
 	ext_os "github.com/open-policy-agent/opa-control-plane/pkg/objectstorage"
 )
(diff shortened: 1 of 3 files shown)


pkg/service/service.go
```diff
@@ -267,6 +275,12 @@ func (s *Service) initDB(ctx context.Context) error {
 	}
 	s.database = *db
 
+	// Re-apply authorizer if one was set before initDB, since migrations.Run()
+	// creates a new database instance with the default authorizer.
```
ashutosh-narkar: What if we add a `WithAuthorizer` on the `Migrator` and pass it that way?

yi-chen-roger: make sense updated

srenatus:
> Also expose the internal logging package via pkg/logging so external consumers can create loggers for WithLogger without importing internal packages.

Since we're touching the logging, I think we should switch to slog.Logger eventually. That is not for _this PR_, but it makes me wonder if we'd do better to expose `WithLogger(slog.Logger)` and introduce/use an adapter in the OCP code (until we've moved that to slog all the way). What do you think? What do you use on your host application, would slog be a good fit?

yi-chen-roger:
> > Also expose the internal logging package via pkg/logging so external consumers can create loggers for WithLogger without importing internal packages.
> 
> Since we're touching the logging, I think we should switch to slog.Logger eventually. That is not for _this PR_, but it makes me wonder if we'd do better to expose `WithLogger(slog.Logger)` and introduce/use an adapter in the OCP code (until we've moved that to slog all the way). What do you think? What do you use on your host application, would slog be a good fit?

make sense. Updated

srenatus (approved):
LGTM Thanks

yi-chen-roger:
> LGTM Thanks
Ty, 

you will help merge the pr, right?

PR #322, opened 2026-04-10
Apply formatting to refactored modules during bundle build

### LINKED ISSUE #321: `opactl build` reformats rego in a way that produces code which fails OPA run/eval
I have a library that contains this comprehension:

```rego
flatten_cpu(cpu_all) = cpu_all_flattened {
    cpu_str := [ replace(cpu, "m", "") | not to_number(cpu_all[i]); cpu := cpu_all[i]]
    cpu_int := [ to_number(cpu) | to_number(cpu_all[i]); cpu := cpu_all[i]]
    cpu_str_int = [ (to_number(cpu_str[i])/1000) | cpu_str[i]]
    cpu_all_flattened := array.concat(cpu_int, cpu_str_int)
}
```

When this library file is pulled in as part of a stacks requirements:

```yaml
requirements:
    - source: kubernetes:v2-stack
    - source: libraries
```
```yaml
sources:
  libraries:
    name: libraries
    git:
      repo: https://github.com/...
      reference: refs/heads/master
      included_files:
      - libraries/*
      excluded_files:
      - .*/*
      credentials: workspace_git
```

opactl build transforms the code as follows (in the resulting bundle):

```rego
flatten_cpu(cpu_all) = cpu_all_flattened {
	cpu_str := [replace(cpu, "m", "") | not to_number(cpu_all[i]); cpu := cpu_all[i]]
	cpu_int := [to_number(cpu) | to_number(cpu_all[i]); cpu := cpu_all[i]]
	cpu_str_int = [to_number(cpu_str[i]) / 1000 | cpu_str[i]]
	cpu_all_flattened := array.concat(cpu_int, cpu_str_int)
}
```

You can see the parenthesis were removed on this line `cpu_str_int = [to_number(cpu_str[i]) / 1000 | cpu_str[i]]`

This causes previously functional code to report the following:

```
policy.rego:16: rego_unsafe_var_error: var i is unsafe
policy.rego:18: rego_unsafe_var_error: var cpu_str_int is unsafe
policy.rego:18: rego_unsafe_var_error: var cpu_str is unsafe
policy.rego:19: rego_unsafe_var_error: var cpu_all_flattened is unsafe
policy.rego:19: rego_unsafe_var_error: var cpu_int is unsafe
```

I ran this through Claude and asked it to find the source of the bug in opa-control-plane and it reported the following. I verified by changing both the bundle and/or the source code according and it does seem to fix the issue:

```
The bug is in [pkg/builder/builder.go](file:///Users/rahners/sandbox/opa-control-plane/pkg/builder/builder.go) in the `extractAndTransformRego()` function. Let me show you the exact code:   

Read [](file:///Users/rahners/sandbox/opa-control-plane/pkg/builder/builder.go)

There's the bug — **line 643**: `rendered[p] = m.String()`

Here's the flow:
1. `.rego` files are parsed into `*ast.Module` objects (line 615)
2. Package paths are refactored via `refactor.New().Move()` (line 636)
3. Modules are rendered back to source with `m.String()` (line 643)

**`ast.Module.String()` uses OPA's AST printer, which strips parentheses it considers "unnecessary"** — but the parens in `(to_number(cpu_str[i])/1000)` are semantically important inside the list comprehension because they group the output expression before the `|` pipe.

The fix would be to use `format.Ast(m)` instead of `m.String()`, or better yet, apply the package path transformation at the source text level rather than round-tripping through the AST. The bug is upstream in OPA's `ast.Module.String()` method, but the control plane triggers it by doing the parse → transform → re-render cycle.
```

johanfylling:
instead of using `.String()` function on `ast.Module` which isn't guaranteed to always generate proper Rego.

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -22,6 +22,7 @@ import (
 	"github.com/open-policy-agent/opa/bundle"  // nolint:staticcheck
 	"github.com/open-policy-agent/opa/compile" // nolint:staticcheck
 	"github.com/open-policy-agent/opa/rego"    // nolint:staticcheck
+	"github.com/open-policy-agent/opa/v1/format"
 	"github.com/open-policy-agent/opa/v1/refactor"
 	"github.com/open-policy-agent/opa/v1/topdown"
 
@@ -642,7 +643,11 @@ func extractAndTransformRego(fsys fs.FS, mnts []mount) (fs.FS, error) {
 
 	rendered := make(map[string]string, len(modules))
 	for p, m := range modules {
-		rendered[p] = m.String()
+		r, err := format.Ast(m)
+		if err != nil {
+			return nil, fmt.Errorf("failed to format module %s, %w", p, err)
+		}
+		rendered[p] = string(r)
 	}
 
 	return ocp_fs.MapFS(rendered), nil
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -648,7 +648,6 @@ func extractAndTransformRego(fsys fs.FS, mnts []mount) (fs.FS, error) {
 			return nil, fmt.Errorf("failed to format module %s, %w", p, err)
 		}
 		rendered[p] = string(r)
-		//rendered[p] = m.String()
 	}
 
 	return ocp_fs.MapFS(rendered), nil
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -21,8 +21,8 @@ import (
 	"github.com/open-policy-agent/opa/ast"     // nolint:staticcheck
 	"github.com/open-policy-agent/opa/bundle"  // nolint:staticcheck
 	"github.com/open-policy-agent/opa/compile" // nolint:staticcheck
+	"github.com/open-policy-agent/opa/format"  // nolint:staticcheck
 	"github.com/open-policy-agent/opa/rego"    // nolint:staticcheck
-	"github.com/open-policy-agent/opa/v1/format"
 	"github.com/open-policy-agent/opa/v1/refactor"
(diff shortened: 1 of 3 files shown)


srenatus (approved):
LGTM, but I kind of thought that .String() would always yield valid, if badly-formatted, Rego

pkg/builder/builder.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/open-policy-agent/opa/bundle"  // nolint:staticcheck
 	"github.com/open-policy-agent/opa/compile" // nolint:staticcheck
 	"github.com/open-policy-agent/opa/rego"    // nolint:staticcheck
+	"github.com/open-policy-agent/opa/v1/format"
```
johanfylling: I'm a bit concerned about the mix of v0 and v1 imports here. Which Rego version is the intended outcome? We rely on this being supplied through the source bundles? But sources aren't necessarily bundles, are they? This isn't something users can control through config?

If the intention is that the baseline Rego version is expected to be v0, which users can then affect through the `rego.v1` import, then we should be importing `"github.com/open-policy-agent/opa/format"` here (or supply the version through formatting options).

johanfylling:
> but I kind of thought that .String() would always yield valid, if badly-formatted, 

Yeah, that's the common wisdom. But apparently not for parenthesis wrapping compensation terms 😅. We should of course fix this in OPA, but I figured we could start making this end a bit more robust.

pkg/builder/builder.go
srenatus: I think it’s consistent to use the v0 import here. Something to eventually tackle, but not here and now.

srenatus (approved):

PR #338, opened 2026-05-04
Refactor ObjectStorage's Upload to use options struct

ashutosh-narkar:
This change replaces positional parameters (tenant, name, revision, totalSize) with
a single options struct. This makes the interface easier to
extend without breaking existing callers and new options can be added
without changing the Upload signature.

### DIFF
--- a/pkg/objectstorage/objectstorage.go
+++ b/pkg/objectstorage/objectstorage.go
@@ -10,13 +10,21 @@ import (
 // is stored and no upload was performed.
 var ErrNotModified = errors.New("object not modified")
 
+// UploadOptions holds parameters for an Upload call.
+type UploadOptions struct {
+	Tenant    string
+	Name      string
+	Revision  string
+	TotalSize int64
+}
+
 // ObjectStorage defines the interface for uploading and downloading bundle artifacts
 // to/from object storage systems (e.g., S3, GCS, Azure Blob Storage).
 type ObjectStorage interface {
 	// Upload stores a bundle artifact in object storage.
 	// Implementations may return ErrNotModified to indicate that the upload
 	// was skipped because the content has not changed.
-	Upload(ctx context.Context, body io.ReadSeeker, name string, revision string, totalSize int64) error
+	Upload(ctx context.Context, body io.ReadSeeker, opts UploadOptions) error
 
 	// Download retrieves a bundle artifact from object storage.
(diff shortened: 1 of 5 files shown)


pkg/objectstorage/objectstorage.go
```diff
@@ -16,7 +16,7 @@ type ObjectStorage interface {
 	// Upload stores a bundle artifact in object storage.
 	// Implementations may return ErrNotModified to indicate that the upload
 	// was skipped because the content has not changed.
-	Upload(ctx context.Context, body io.ReadSeeker, name string, revision string, totalSize int64) error
+	Upload(ctx context.Context, body io.ReadSeeker, tenant, name string, revision string, totalSize int64) error
```
srenatus: If we need to mess this up, which is OK for now, would you mind introducing  functional options? SO we'd do
```go
storage.Upload(ctx,
  WithReader(body),
  WithTenant(tenant),
  WithName(name),
  WithRevision(Revision),
  WithTotalSize(size))
```

....so we wouldn't break the function signature if we add another field later on. The "With..." is just something I came up with on the spot, there probably is a more ergonomic way 😉

ashutosh-narkar: Makes sense. Updated to use functional options pattern.

pkg/objectstorage/objectstorage.go
```diff
@@ -10,13 +10,46 @@ import (
...
+}
+
 // ObjectStorage defines the interface for uploading and downloading bundle artifacts
 // to/from object storage systems (e.g., S3, GCS, Azure Blob Storage).
 type ObjectStorage interface {
 	// Upload stores a bundle artifact in object storage.
 	// Implementations may return ErrNotModified to indicate that the upload
 	// was skipped because the content has not changed.
-	Upload(ctx context.Context, body io.ReadSeeker, name string, revision string, totalSize int64) error
+	Upload(ctx context.Context, body io.ReadSeeker, opts ...UploadOption) error
```
srenatus: Hm! With the fields (`o.Name`, `o.Revision`, ...)  exposed, there's no difference to just taking an options struct here. 🤔 What do you think, what's preferable?

ashutosh-narkar: I guess in this case we can just pass the options struct. It's probably cleaner. Updated the code.

srenatus (approved):

Issue #170, opened 2025-11-03
expose bundle build args


srenatus:
Currently, there is no way to influence how a bundle gets build, like you can with `opa build`. Optimization levels are a something that people might want to control. If O=1 and O=2 are exposed, we'll also need to expose the inlining control settings. It should also be possible to build plan or wasm bundles.

It could be a new `options` field that's per-bundle, in both the config and the HTTP handlers.

📝 Laundry list
- [x] optimization #282 
- [ ] v0/v1
- [ ] capabilities
- [x] revision #268 
- [x] target (plan/IR, wasm, rego) #220

PR #340, opened 2026-05-07
Respect source .manifest when building bundles

charlieegan3:
Read rego_version, roots, and revision from source .manifest files so v1 Rego syntax compiles correctly without requiring import rego.v1.

### DIFF
--- a/e2e/cli/build_v1_rego.txtar
+++ b/e2e/cli/build_v1_rego.txtar
@@ -1,0 +1,31 @@
+# Regression test: a source directory with a .manifest declaring rego_version:1
+# and v1 Rego syntax (partial rules using 'contains') but no 'import rego.v1'
+# should build successfully.
+exec $OPACTL build --config config.d/bundle.yml --data-dir tmp --non-interactive
+stderr 'source manifest roots .* differ from computed roots .* using manifest roots'
+stderr 'source manifest revision "foobar" overrides WithRevisionFunc result; using manifest revision'
+
+exec tar xf bundles/my-policies/bundle.tar.gz
+exec jq -e '.rego_version == 1 and .revision == "foobar" and .roots == ["rules","foobar"]' .manifest
+
+-- files/sources/my-policies/.manifest --
+{"rego_version":1,"revision":"foobar","roots":["rules","foobar"]}
+-- files/sources/my-policies/rules/rules.rego --
+package rules
+
+rejection_reasons contains msg if {
+    msg := "test reason"
+}
+-- config.d/bundle.yml --
+bundles:
+  my-policies:
+    object_storage:
+      filesystem:
+        path: bundles/my-policies/bundle.tar.gz
+    requirements:
+    - source: my-policies
+sources:
+  my-policies:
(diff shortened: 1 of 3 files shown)


e2e/cli/build_v1_rego.txtar
```diff
@@ -0,0 +1,31 @@
+# Regression test: a source directory with a .manifest declaring rego_version:1
+# and v1 Rego syntax (partial rules using 'contains') but no 'import rego.v1'
+# should build successfully.
+exec $OPACTL build --config config.d/bundle.yml --data-dir tmp --non-interactive
+stderr 'source manifest roots .* differ from computed roots .* using manifest roots'
+stderr 'source manifest revision "foobar" overrides WithRevisionFunc result; using manifest revision'
+
+exec tar xf bundles/my-policies/bundle.tar.gz
+exec jq -e '.rego_version == 1 and .revision == "foobar" and .roots == ["rules","foobar"]' .manifest
```
srenatus, line 9: Huh, we don't have the `jq` built-in here yet. 😓 So nothing to comment, this is fine!

srenatus: I suppose e could also assert the complete `.manifest`, since it's going to be pretty static? Like `cmp .manifest exp/manifest` with some file block `-- exp/manifest --` to be added below. But it doesn't matter, this is fine, and more lenient if we happen to add something unrelated later.

srenatus (approved):
Hah turns out there is nothing to nitpick here! Thanks a lot! 👏

srenatus:
Cool new feature! 🎈 Thanks for working on this.

Respecting `.manifest` lets us
1. accept arbitrary extra information (we sometimes abuse the manifest for config data 💦)
2. control the rego version (relevant alternative to #170 for that)
3. externally control the revision of a bundle

e2e/cli/build_v1_rego.txtar
charlieegan3: I went with jq since the file is being re-written by the build and didn't want json discrepancies in just the bytes.

charlieegan3:
Thanks for the review Stephan! and the guidance

pkg/builder/builder.go
```diff
@@ -320,10 +323,21 @@ func (b *Builder) Build(ctx context.Context) error {
 				return err
 			}
 
+			regoVersion := ast.RegoV0
+			if m, ok := readManifest(fs0); ok {
```
gusega, line 327: @charlieegan3 But manifest is per bundle, right? A bundle may contain multiple sources. I find it a bit counterintiutive that each source could provide a rego version configuration for a part of the bundle. Would it make sense to instead provide the bundle manifest configuration on the bundle level in the `config.yaml`:

```yaml
bundles:
  my-bundle:
    manifest: 
      ## manifest properties
    requirements:
      ## ...
```

charlieegan3: Hey, I just saw this comment. Sorry about that, I'm a little unclear on the bundle having multiple sources, it might be best to open an issue for the use case there so we have it tracked.

PR #343, opened 2026-05-13
pkg: expose typed database config API

yi-chen-roger:
## Summary

  Add `DatabaseConfig`, `SQLDatabaseConfig`, and `AmazonRDSConfig` public types to `pkg/config` so callers can build typed database configuration instead of marshaling raw JSON/YAML bytes.

  Add `InitDBWithConfig` to `pkg/database.Database` and `WithDatabaseConfig` to `pkg/service.Service` as type-safe alternatives to `InitDB`/`WithRawConfig` for database-only configuration. Both existing methods are kept
  for backward compatibility.

  ## Changes

  - `pkg/config`: new `DatabaseConfig`, `SQLDatabaseConfig`, `AmazonRDSConfig` types
  - `pkg/database`: add `InitDBWithConfig(ctx, *config.DatabaseConfig)`, deprecate `InitDB`
  - `pkg/service`: add `WithDatabaseConfig(*config.DatabaseConfig)` alongside existing `WithRawConfig`

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -8,8 +8,9 @@
 //
 //	db := database.New()
 //	db.WithAuthorizer(myAuthorizer)
-//	rawConfig := []byte(`{"database": {"sql": {"driver": "sqlite3", "dsn": "file::memory:?cache=shared"}}}`)
-//	if err := db.InitDB(ctx, rawConfig); err != nil {
+//	if err := db.InitDBWithConfig(ctx, &config.DatabaseConfig{
+//	    SQL: &config.SQLDatabaseConfig{Driver: "sqlite3", DSN: "file::memory:?cache=shared"},
+//	}); err != nil {
 //	    log.Fatal(err)
 //	}
 //	defer db.CloseDB()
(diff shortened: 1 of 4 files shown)


srenatus:
Tests don't look happy. I can review this in depth next week, but @ashutosh-narkar can, too, I would think. At this point, I wonder if there's any reason to keep a distinction between internal and external config at all. 🤔 What do you think?

ashutosh-narkar:
>  I wonder if there's any reason to keep a distinction between internal and external config at all

At this point, most of the structs are aliases to the public package anyways.

yi-chen-roger:
@srenatus ptal

pkg/database/database.go
```diff
@@ -78,15 +80,18 @@ func (d *Database) WithAccessFactory(af ext_authz.AccessFactory) *Database {
 
 // InitDB initializes the database connection from a raw root configuration.
 //
-// The rawConfig must be a JSON (or YAML) document containing a "database" key.
-// Example:
-//
-//	{"database": {"sql": {"driver": "sqlite3", "dsn": "file::memory:?cache=shared"}}}
+// Deprecated: prefer InitDBWithConfig for type-safe configuration.
```
srenatus, line 83: Note: When we actually want to deprecate them in a future release the way forward would be to advise any users of this to json.Unmarshal their rawConfig into `config.DatabaseConfig`. Simple enough that we can get rid of this eventually.

srenatus (approved):
Thanks!

PR #344, opened 2026-05-13
database: wrap driver errors in domain sentinel errors

yi-chen-roger:
OCP store layer now translates pgconn.PgError (PostgreSQL/CockroachDB), mysql.MySQLError, and SQLite constraint errors into ErrAlreadyExists and ErrConflict before returning to callers, so upstream consumers can use errors.Is() without importing database driver packages.

- Add ErrAlreadyExists and ErrConflict sentinel errors
- Add translateStoreError() called at all upsertReturning return paths
- Update errorAuto() to use errors.Is() and handle new error types
- Add CodeAlreadyExists and CodeConflict HTTP error codes
- Re-export new sentinel errors from pkg/database


srenatus (approved):
LGTM but let's get this green. PTAL

srenatus (approved):

PR #347, opened 2026-05-18
database: add NewFromDB constructor and transactional tenant+principal upset

yi-chen-roger:
Adds two new capabilities to enable callers to share a single database connection and participate in caller-managed transactions:

- internal/database: NewFromDB creates a Database from an existing *sql.DB and driver name, avoiding the need for a separate connection pool.
- internal/database: UpsertTenantAndPrincipalTx accepts a *sql.Tx so callers can include tenant/principal creation within their own transaction. UpsertTenantAndPrincipal is refactored to delegate to the Tx variant.
- pkg/database: expose both as NewFromDB and UpsertTenantWithPrincipalTx.


internal/database/database.go
```diff
@@ -66,6 +66,35 @@ func New() *Database {
...
+		db:            db,
+		executeTx:     executeTx,
+		authorizer:    &authz.OPAuthorizer{},
+		accessFactory: authz.NewAccess,
+	}
+	switch driver {
+	case "sqlite3", "sqlite":
+		d.kind = sqlite
+	case "postgres", "pgx":
+		d.kind = postgres
```
srenatus:
```
--- FAIL: TestNewFromDB (0.00s)
    --- FAIL: TestNewFromDB/postgres (3.02s)
        database_test.go:1038: NewFromDB: unsupported driver: postgresql
FAIL
```

srenatus: Makes me wonder if we could get the proper driver name from `*sql.DB`... 🤔

srenatus (commented):
Code looks good but CI doesn't. Please adjust the string driver names vs `d.kind` in `NewFromDB`.

internal/database/database.go
```diff
@@ -66,6 +66,35 @@ func New() *Database {
...
+	case "postgres", "pgx", "postgresql":
+		d.kind = postgres
+	case "cockroachdb":
+		d.kind = cockroach
+		d.executeTx = crdb.ExecuteTx
+	case "mysql":
+		d.kind = mysql
+	default:
+		return nil, fmt.Errorf("unsupported driver: %s", driver)
+	}
```
srenatus, line 94: For future reference: I wanted to avoid having the extra `driver` parameter and explored using type assertions on `db.Driver()` to figure out the kind of SQL db we're talking to. It _almost_ worked -- but it failed with cockroachdb: We're using pgx for that, and pgx cannot tell us if it's set up to talk to a CRDB or plain PG. We'd need to open a connection and inspect the crdb_version parameter. That's when I decided that `driver string` is the better trade-off. 😓

srenatus (approved):

internal/database/database.go
srenatus: I think the better question is if you're using the same ExecuteTx wrapper for CRDB as OCP is (https://pkg.go.dev/github.com/cockroachdb/cockroach-go/v2/crdb#ExecuteTx). If you happen to use something more complex, we might need to allow passing _that_ as a second parameter. 🤔

PR #349, opened 2026-05-20
metrics: configurable metrics

bdumpp:
This solves #292 and enables configuration of metrics. Metrics are enabled by default and can be disabled globally, per package or per per metric by using the proper of the following settings in your config file
```
metrics.enabled: false
metrics.<package>.enabled: false
metrics.<package>.<metric>.enabled: false
```
Buckets for histograms can be changed by setting
```
metrics.<package>.<metric>.BucketsSeconds: [1, 2, 3, 4]
```


internal/metrics/init.go
```diff
@@ -0,0 +1,24 @@
...
+// Init initializes all metrics collectors with the given configuration.
+// Must be called once before any metrics are recorded.
+func Init(cfg *config.MetricsConfig) {
+	initHTTPMetrics(cfg)
+	initGitSyncMetrics(cfg)
+	initWorkerMetrics(cfg)
+}
+
+func isEnabled(enabled *bool) bool {
+	return enabled == nil || *enabled
```
srenatus, line 37: So that's to say "enabled by default" or "enabled if unset", right? ✔️

srenatus (commented):
Thanks for contributing, this looks good so far!

Test failures might mean that we're using some global registry? I guess we used that before, if but if what had to change to make this work, so be it. I'd imagine that we make `Init()` take the registry, and just pass the default registry in the service startup, whereas we use a test-scoped one for tests. 🤔

internal/metrics/init.go
bdumpp: yes

cmd/run/run.go
```diff
@@ -71,7 +72,12 @@ func init() {
 			}
 
 			go func() {
-				if err := server.New().WithDatabase(svc.Database()).WithReadiness(svc.Ready).WithConfig(config.Service).Init().ListenAndServe(params.addr); err != nil {
+				if err := server.New().
+					WithDatabase(svc.Database()).
+					WithReadiness(svc.Ready).
+					WithConfig(config).
+					Init(prometheus.DefaultRegisterer).
```
srenatus: Hmm looking at this, I think we should go with a slight change,

```suggestion
				if err := server.New().
					WithDatabase(svc.Database()).
					WithReadiness(svc.Ready).
					WithConfig(config).
					WithPrometheusRegisterer(prometheus.DefaultRegisterer).
					Init().
```
and `Init()` will check if the prom reg is != nil, register accordingly.

config/schema.go
```diff
@@ -1,4 +1,5 @@
 //go:generate go run ../build/gen-config-schema.go schema.json
+//go:generate go run ../build/gen-config-schema.go ../schema.json
```
srenatus, line 2: What's this about? Do we need both targets? Have we just forgotten to run this in a while...? 🤔

srenatus (approved):
Thank you! 🎈 🎉

config/schema.go
bdumpp: I don't know why we have two schema.json files, but while debugging a typo I was looking at the wrong one and not understanding why my changes did not change the schema.

PR #360, opened 2026-06-09
feat(gcs): skip upload if bundle is unmodified  

taurelius:
Updates GCS implementation to skip bundle upload when an identical bundle is generated, aligning with the S3 and FileSystem behaviour.


srenatus (approved):
Solid work, thanks for contributing! 👏 

Have you tried this with real GCP CS?

srenatus:
Ah, sorry, I've just merged a dependency bump. Would you mind rebasing and solving the conflict? Thanks!

taurelius:
@srenatus no worries, conflicts resolved 😌 

And yes - tested against GCS, working as expected!

PR #363, opened 2026-06-16
database: translate missing required row to ErrConflict in lookups

yi-chen-roger:
lookupID returns the raw sql.ErrNoRows when a SELECT misses, which is fine for the lookups whose callers explicitly handle that case (e.g. optional secret resolution). For lookups where the caller's input must reference an existing row — bundle/source/stack requirements naming an existing source — a missing row is conceptually a foreign-key violation, just enforced at the application layer instead of by the DB. Today such errors leak as 'sql: no rows in result set', and gRPC layers above that have no sentinel to map and surface them as Internal.

Add lookupRequiredID alongside lookupID; it wraps sql.ErrNoRows as ErrConflict (consistent with translateStoreError's mapping for the foreign-key 23503 case) with a clear message identifying the missing row. Use it at the three requirement-source call sites in UpsertBundle, UpsertSource (requirements) and UpsertStack.


srenatus:
> For lookups where the caller's input must reference an existing row — bundle/source/stack requirements naming an existing source — a missing row is conceptually a foreign-key violation, just enforced at the application layer instead of by the DB.

I follow your reasoning this far, but going from FK violation to _conflict error_ is a step I'm struggling with. Maybe the translateStoreError mapping is weird? 🤔

srenatus (approved):
The error question aside the code looks good to me. 👍

yi-chen-roger:
> > For lookups where the caller's input must reference an existing row — bundle/source/stack requirements naming an existing source — a missing row is conceptually a foreign-key violation, just enforced at the application layer instead of by the DB.
> 
> I follow your reasoning this far, but going from FK violation to _conflict error_ is a step I'm struggling with. Maybe the translateStoreError mapping is weird? 🤔

https://github.com/open-policy-agent/opa-control-plane/blob/fd74f90d6fd42436e477c8780619a7515c4a0125/internal/database/errors.go#L18 
mentioned // ErrConflict is returned when an operation conflicts with the current state

Yes — these aren't just conceptually equivalent, they're literally enforcing the same constraint. The schema declares bundles_requirements.source_id, sources_requirements.requirement_id, and stacks_requirements.source_id all as FKs to sources(id). lookupID at these three sites exists solely to translate the user-supplied name into the id that will be inserted  into those FK-constrained columns. If we skipped the lookup and inserted with a non-existent id, the DB would fire 23503 — which translateStoreError already maps to ErrConflict.

srenatus:
> If we skipped the lookup and inserted with a non-existent id, the DB would fire 23503 — which translateStoreError already maps to ErrConflict.

That's the bit I'm wondering about: is it really a good error translation to turn a "wrong name of a thing" error into a 
"conflict"? It's not conflicting -- it's not there.

Issue #365, opened 2026-06-24
Add an option to reduce OPA memory pressure when loading large data.json files


Pushpalanka:
### Context:
When a system source has multiple HTTP datasources, OCP collapses all data into a single root /data.json in the bundle tar. OPA agents must parse and hold the entire merged document in memory at once, causing memory pressure on systems with large datasources. When loading the bundle with a single large data.json, GC doesn't get to clean the whole large parsed file until variables are released, where as when data is distributed among multiple small data.json files GC could act faster.

### Suggestion:
Introduce a per-bundle flag options.nested_data_files (default: false). When enabled, the bundle builder writes each datasource's data as a separate data.json at its namespaced path within the tar (e.g. /teams/data.json, /admins/data.json), matching the previous Styra DAS bundle layout. Since OPA's bundle reader merges nested files as seen with previous structure the runtime semantics are expected to be unchanged.

The flag is per-bundle because OCP can be building bundles for many systems with different data characteristics: systems with many small datasources may not benefit from this and would carry extra tar-entry overhead.

PR #366, opened 2026-06-24
Add an option to reduce OPA agent memory pressure when loading large data.json files

Pushpalanka:
Relates: https://github.com/open-policy-agent/opa-control-plane/issues/365

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -1,8 +1,10 @@
 package builder
 
 import (
+	"archive/tar"
 	"bytes"
 	"cmp"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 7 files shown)


pkg/builder/builder.go
```diff
@@ -499,9 +508,118 @@ func (b *Builder) Build(ctx context.Context) error {
 	result.Manifest.SetRegoVersion(effectiveRegoVersion)
 	result.Manifest.Revision = b.revision
 
+	if b.nestedDataFiles {
+		return writeBundle(b.output, *result, bsFSMap)
+	}
 	return bundle.Write(b.output, *result)
```
srenatus, line 514: 🏗️ I think this feature would make a lot of sense as an option on `bundle`. We wouldn't have to replicate the tarball creation logic here, and any OPA users would benefit from it, and also OCP.

What do you think? Would be you OK with pivoting into that direction? 😃

Pushpalanka: Thanks for the direction and taking time to check the draft @srenatus 
That makes lot of sense, will try that path. 👍

srenatus: Thank you! And thanks for contributing, this looks like such a nice tweak. ✨

PR #368, opened 2026-06-30
pool: add Stop() to allow worker goroutines to exit

yi-chen-roger:
Worker goroutines in Pool.work() block indefinitely on dequeue() when the task queue is empty (waiting up to 1 year for a fake deadline). When the pool is used in single-shot mode — where New() is called on every run — these goroutines accumulate and are never reclaimed.

Add a done channel to Pool. dequeue() now selects on p.done in addition to the timer and wait channel, returning (nil, false) when Stop() is called. work() exits cleanly when dequeue() signals shutdown.

Service.Run() defers pool.Stop() so all worker goroutines exit when Run() returns, fixing the leak in single-shot usage.


internal/pool/pool.go
```diff
@@ -95,6 +108,9 @@ func (p *Pool) dequeue() *task {
 			select {
 			case <-time.After(time.Until(t.deadline)):
 			case <-wait:
+			case <-p.done:
+				p.mu.Lock()
```
srenatus, line 112: This line is the only surprise here. Why do we lock before the return?

yi-chen-roger: updated

srenatus:
PTAL at the failing tests

internal/pool/pool.go
yi-chen-roger: have to add the p.mu.Lock() back to fix  the ci fails, we need to lock it so later the dequeue defer unlock have a lock to unlock after return

PR #373, opened 2026-07-13
pool: propagate task context instead of context.Background()

yi-chen-roger:
## What
`Pool.work()` executed every task with a fresh `context.Background()`, discarding whatever context the caller passed to `Add`. Callers relying on context cancellation to bound in-flight work (e.g. a retry loop selecting on `ctx.Done()`) never actually observe it, since the pool hands the task a context that is never cancelled.

## Changes
- `Pool.Add` now takes an explicit `context.Context`, stored on the task and passed to every invocation of `fn`, including rescheduled executions (when `fn` returns a future deadline). `Service.launchWorkers` passes its `ctx` through to `Pool.Add` so `BundleWorker.Execute` (and anything it calls) sees the real, cancellable context instead of a detached background one.
- Fixed the single-shot polling loop in `Service.Run`, which busy-waited on `allWorkersDone()` without ever checking `ctx.Done()`, so a caller-supplied timeout had no effect in single-shot mode.

## Testing
- `go test ./internal/pool/...` — added `TestPoolPropagatesTaskContext` and `TestPoolObservesContextCancellation` to cover the new behavior.
- `go test ./pkg/service/...` (network-bound httptest subtests excluded in this sandbox, unaffected by this change; verified pre-existing on main too).


internal/pool/pool_test.go
```diff
@@ -8,19 +8,20 @@ import (
 
 func TestPool(t *testing.T) {
 	p := New(2)
+	ctx := context.Background()
 
 	// Add a task that returns a deadline in the future.
-	p.Add(func(_ context.Context) time.Time {
+	p.Add(ctx, func(_ context.Context) time.Time {
```
srenatus: [nit] Also below
```suggestion
	p.Add(ctx, func(context.Context) time.Time {
```

srenatus (approved):
LGTM if the tests pass 😉 They're still running at the moment.

internal/pool/pool_test.go
yi-chen-roger: updated

PR #374, opened 2026-07-16
service: use CockroachDB follower reads for periodic worker list refresh

yi-chen-roger:
## Why
`launchWorkers` periodically re-lists all bundles/sources/stacks for every tenant via broad multi-table LEFT JOIN queries (`ListBundles`, `ListSources`, `ListStacks`). These joins hold a read transaction open long enough, over the same tables that `Upsert*`/`Delete*` write to, that under load they see frequent CockroachDB serialization-conflict retries (each retry re-running the whole join, compounding latency/CPU).

## What
- `internal/database/database.go`: new `ListOptions.Stale` flag + `applyFollowerRead` helper that runs `SET TRANSACTION AS OF SYSTEM TIME follower_read_timestamp()` as the first statement in the transaction when `Stale=true` and the dialect is CockroachDB. No-op on SQLite/Postgres/MySQL, and no-op when `Stale` isn't set — other callers of `ListBundles`/`ListSources`/`ListStacks` that need fresh reads are unaffected.
- `pkg/service/service.go`: `launchWorkers`'s three periodic list calls now pass `Stale: true`. This bookkeeping read doesn't need up-to-the-moment consistency, so it's safe to read a few-seconds-stale snapshot that can't conflict with concurrently committing writes.

## Out of scope
The underlying queries are also each a single, wide multi-table JOIN with row fan-out (bundle x secrets x requirements) and no pagination limit in this call path — a separate, more invasive fix (splitting into flat batched queries) is left as a follow-up. This PR only addresses the contention/retry symptom for the periodic worker refresh path.

## Testing
- `go build ./...` — clean.
- `go vet ./internal/database/... ./pkg/service/...` — clean.
- `go test ./internal/database/...` (SQLite-backed subtests, which exercise `ListBundles`/`ListSources`/`ListStacks`) — pass.
- `gofmt -l` — clean on both changed files.

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -118,6 +118,14 @@ type ListOptions struct {
 	Limit  int
 	Cursor string
 	name   string
+
+	// Stale allows the list query to read a slightly stale (a few seconds
+	// old) snapshot via CockroachDB follower reads instead of the current
+	// timestamp. Use this for internal bookkeeping reads (e.g. periodic
+	// worker refresh) that don't need up-to-the-moment consistency and
+	// would otherwise contend with concurrent writes to the same tables.
+	// No-op on non-CockroachDB dialects.
+	Stale bool
 }
 
 func (opts ListOptions) cursor() int64 {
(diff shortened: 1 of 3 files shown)


srenatus:
LGTM if this is what people commonly do with CRDB. Please fix DCO and have a look at the failing tests when you find the time! Thanks

srenatus (approved):
Ok to merge when green. Thanks!

srenatus:
Sorry I had forgotten about this one. You still care?

yi-chen-roger:
> Sorry I had forgotten about this one. You still care?

yes, I still want to merge this

ashutosh-narkar (approved):

PR #375, opened 2026-07-17
Persist sync/transform failures in bundle status table

ashutosh-narkar:
Bundle builds run in four phases (sync, transform, build, push), but only build and push outcomes were persisted to the bundles_statuses table. Sync/transform failures were logged so operators querying bundle status could never see them. This change includes sync/transform failures in the bundle status table.

Behavior change: bundles without a configured revision previously produced no status rows at all so GetLatestBundleStatus returned ErrNotFound and the status API returned 404 for them. They now
persist a single status row keyed on the empty-string sentinel, updated in place each tick to reflect the current phase/state, so those queries now return a status instead of not-found. Bundles that always resolve a non-empty revision are unaffected.

### DIFF
--- a/internal/database/bundle_status.go
+++ b/internal/database/bundle_status.go
@@ -11,15 +11,26 @@ import (
 
 const MaxBundleStatusRetention = 10
 
+// SentinelRevision keys the single "pre-revision state" row per bundle: the
+// outcome of failures that happen before a revision is resolved (sync, transform,
+// internal). It is the canonical value callers pass to UpsertBundleStatus for
+// pre-revision phases (e.g. pkg/service's worker). A present sentinel row means
+// the bundle is currently failing before the build phase (it is deleted as soon
+// as a real-revision row is written; see UpsertBundleStatus).
+const SentinelRevision = ""
+
 // UpsertBundleStatus creates or updates a bundle status record with the given phase and status.
 // For a given tenant+bundle+revision combination, only one record exists.
 // If a record already exists for the combination, it updates the phase and status.
+//
+// A revision of SentinelRevision (the empty string) records a pre-revision phase
+// (sync/transform/internal) and dedups to one sentinel row per bundle. When a
+// real (non-sentinel) revision is written, any existing sentinel row for the
+// bundle is deleted in the same transaction, so the sentinel never
+// lingers as a stale failure after the run gets past sync/transform.
+//
 // Old records beyond the retention limit (MaxBundleStatusRetention) are cleaned up in the same transaction.
 func (d *Database) UpsertBundleStatus(ctx context.Context, tenant, bundle, revision, phase, status, errMsg string) (int, error) {
-	if revision == "" {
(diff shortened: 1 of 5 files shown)


pkg/service/worker.go
```diff
@@ -227,21 +227,7 @@ func (w *BundleWorker) Execute(ctx context.Context) time.Time {
...
-		}
-		return w.report(ctx, BuildStateBuildFailed, startTime, err)
-	}
-
-	if b.Revision() != "" {
-		_, err := w.database.UpsertBundleStatus(ctx, w.tenant, w.bundleConfig.Name, b.Revision(), BuildPhaseBuild.String(), BuildStateSuccess.String(), "")
-		if err != nil {
-			w.log.Warnf("failed to track bundle build state %q: %v", w.bundleConfig.Name, err)
-		}
+		return w.report(ctx, BuildStateBuildFailed, BuildPhaseBuild, b.Revision(), startTime, err)
```
srenatus: [nit] I see it was done like that previously, but can we use `resolvedRevision` here instead of `b.Revision()`?

internal/database/bundle_status.go
```diff
@@ -11,15 +11,26 @@ import (
 
 const MaxBundleStatusRetention = 10
 
+// SentinelRevision keys the single "pre-revision state" row per bundle: the
+// outcome of failures that happen before a revision is resolved (sync, transform,
+// internal). It is the canonical value callers pass to UpsertBundleStatus for
+// pre-revision phases (e.g. pkg/service's worker). A present sentinel row means
+// the bundle is currently failing before the build phase (it is deleted as soon
+// as a real-revision row is written; see UpsertBundleStatus).
+const SentinelRevision = ""
```
srenatus, line 20: Should the existing "GetLatestBundleStatus" code somehow respect bundles "currently failing" in its output..? (https://github.com/open-policy-agent/opa-control-plane/blob/main/internal/server/server.go#L253-L271)

srenatus (approved):
LGTM. Claude flagged an issue with "id as proxy for recency" that would trip on the self-heal mechanism, but it was predicated on bundle revision clash or re-use, which shouldn't be a thing 🤞

internal/database/bundle_status.go
ashutosh-narkar: Not sure I understand. Can you please elaborate?

pkg/service/worker.go
ashutosh-narkar: Updated.

internal/database/bundle_status.go
ashutosh-narkar: Ohhh... you mean if a bundle is currently in a failing phase we should return a non-200 response?

ashutosh-narkar: We can track that separately. That would be a backwards compatible change. Also then should we do the same for the list statutes API if the latest status is failing? From a consumer pov, they could make that call for themselves and just expect the API to give the latest status or list of statuses. WDYT?

PR #377, opened 2026-07-22
service: add WithPrometheusRegisterer for callers without internal/metrics access

yi-chen-roger:
## Summary
`Service.WithMetrics` already accepts a `*metrics.Metrics`, but that type lives in `internal/metrics` and is unreachable from outside this module (Go enforces `internal/` package visibility). External callers that only have a `prometheus.Registerer` (e.g. a consuming service that already builds its own registry) have no way to opt into OCP's built-in git sync, HTTP, and bundle build metrics.

This adds a small convenience method:

```go
func (s *Service) WithPrometheusRegisterer(reg prometheus.Registerer) *Service {
	return s.WithMetrics(metrics.Init(nil, reg))
}
```

It wraps `metrics.Init` with a nil `MetricsConfig`, which enables all metrics with default histogram buckets. Callers that need finer-grained per-metric enable/disable or custom buckets should keep using `WithMetrics` directly (not reachable externally without vendoring `internal/config`, but that's an existing constraint, not something this PR changes).

## Test plan
- Added `TestWithPrometheusRegisterer`, which runs a real (failing, unconfigured-secret) git sync through `Service` with a custom `prometheus.Registry` and asserts `ocp_git_sync_count_total` is present after `Gather()`, proving the wiring reaches `gitsync.Synchronizer`'s metrics recording.
- `go test ./pkg/service/...` passes for this test and all other non-network-bound tests in the package.

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -20,6 +20,7 @@ import (
 
 	pkgsync "github.com/open-policy-agent/opa-control-plane/pkg/sync"
 	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/prometheus/client_golang/prometheus"
 	_ "modernc.org/sqlite"
 
 	"github.com/open-policy-agent/opa-control-plane/internal/config"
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
@yi-chen-roger [this](https://github.com/open-policy-agent/opa-control-plane/pull/378) change, exposes the metrics package and is more configurable (for ex. allowing callers to choose the metric name prefix). Do we still need this?

yi-chen-roger:
Yep, #378 covers this properly (public type + configurable namespace, which is strictly better than what I was proposing here). Closing in favor of that.

PR #378, opened 2026-07-22
metrics: expose collectors as public pkg/metrics with configurable namespace

ashutosh-narkar:
This change moves the Metrics type and collector definitions into a public package (pkg/metrics). This will allow embedders to register OCP's collectors into their own registry and choose the metric name prefix.


srenatus (approved):
There's isn't much left of the internal/metrics package 😅 LGTM!

PR #379, opened 2026-07-22
database: add per-query Prometheus metrics

yi-chen-roger:
## Summary
Adds per-query Prometheus metrics for the `Database` layer (CockroachDB/postgres/mysql), which currently has none. Build-time DB failures today only surface as `BundleBuilderTaskFailureRateHigh{error_type=service}` (in a consumer) with no further detail on which query or operation failed.

New metrics, following the existing `gitsync`/`http`/`worker` pattern in `internal/metrics`:
- `ocp_database_query_count_total{operation, state}` — counter, `state` is `SUCCESS`/`FAILED`
- `ocp_database_query_duration_seconds{operation}` — histogram, `operation` parsed from the SQL statement's first keyword (select/insert/update/...)

## Design

The instrumentation is wired in at the `driver.Connector` level (`internal/database/instrumented.go`), not by touching any of the ~50 methods in `database.go` — every one of them already funnels through `sql.OpenDB(connector)` for postgres/CockroachDB/mysql (sqlite's `sql.Open` path is left alone; it's the in-memory/dev default, not a production case).

The tricky part of this kind of wrapping is `driver.Conn`'s several *optional* interfaces (`QueryerContext`, `ExecerContext`, `Pinger`, `ConnPrepareContext`, `ConnBeginTx`, `SessionResetter`, `NamedValueChecker`, `Validator`) — `database/sql` detects which ones a `Conn` implements to decide its execution path, and a wrapper that doesn't preserve the same set of capabilities silently changes behavior (e.g. falling back to a slower prepare-then-exec path) without erroring.

An earlier version of this PR tried to solve this by constructing a different wrapper type per detected capability combination. `TestInstrumentedConn_ForwardsUnrelatedOptionalInterfaces` caught a real bug in that approach (`driver.Pinger` support silently dropped), which led to redesigning around the pattern used by [`github.com/XSAM/otelsql`](https://github.com/XSAM/otelsql) (not taken as a dependency — it's OTel-native, this stays Prometheus-native to match the rest of `internal/metrics`): the wrapper unconditionally implements every optional interface, and each method checks at call time whether the wrapped conn actually supports it, returning `driver.ErrSkip` when it doesn't — the sentinel `database/sql` defines for exactly this case, so it falls back exactly as it would without any wrapping. This avoids the combinatorial-type problem entirely: there's one wrapper type, and no capability can be silently dropped.

`PrepareContext`/`BeginTx` fallback logic (used when the wrapped conn doesn't implement the optional context-aware version) is adapted from `otelsql`, which itself credits it to `database/sql`'s own internal `ctxutil.go`.

## Wiring
`Database.WithMetrics(*metrics.Metrics)` mirrors the existing `WithLogger`/`WithAuthorizer` pattern. `Service.WithMetrics` now also propagates to `s.database`, mirroring how it already does for `WithLogger`.

Also adds `DatabaseMetrics` to the public `MetricsConfig` schema (`internal/config/config.go`, regenerated `schema.json`/`config/schema.json`) for parity with `GitSyncMetrics`/`WorkerMetrics`/`HTTPMetrics`, so operators can fine-tune or disable it the same way.

## Test plan
- `TestInstrumentedConn_FallsBackWhenUnsupported` — wrapping a conn with neither `QueryerContext` nor `ExecerContext` returns `driver.ErrSkip`, not a panic or a silent no-op.
- `TestInstrumentedConn_ForwardsUnrelatedOptionalInterfaces` — `driver.Pinger` support is preserved through the wrapper (the regression test for the bug the earlier design had).
- `TestInstrumentedConn_RecordsMetrics` — a successful query and a failed exec both show up correctly in `ocp_database_query_count_total` via a real `prometheus.Registry`.
- `TestInstrumentedConn_BeginTxFallback` — the `ctxutil.go`-derived fallback path propagates errors correctly and rejects unsupported non-default isolation levels.
- `go test ./internal/database/... ./internal/metrics/... ./internal/config/...` — all pass except the pre-existing `postgres`/`mysql`/`cockroachdb` testcontainers-based subtests, which fail in my sandbox with "rootless Docker not found" — confirmed this also fails identically on unmodified `main`, so it's an environment limitation, not something this PR broke. Would appreciate a maintainer or CI run to confirm this actually instruments real CockroachDB/postgres/mysql connections end-to-end, since that's the one thing my sandbox couldn't verify directly.

### DIFF
--- a/internal/database/instrumented.go
+++ b/internal/database/instrumented.go
@@ -1,0 +1,208 @@
+package database
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"errors"
+	"time"
+
+	"github.com/open-policy-agent/opa-control-plane/pkg/metrics"
+)
+
+// instrumentedConnector wraps a driver.Connector so every driver.Conn it
+// produces records query metrics. Connect is a straight pass-through; the
+// returned Conn is wrapped by instrumentedConn (see below).
+type instrumentedConnector struct {
+	driver.Connector
+	metrics *metrics.Metrics
+}
+
+func newInstrumentedConnector(inner driver.Connector, m *metrics.Metrics) driver.Connector {
+	return instrumentedConnector{Connector: inner, metrics: m}
+}
+
+func (c instrumentedConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	conn, err := c.Connector.Connect(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return wrapConn(conn, c.metrics), nil
+}
+
+// instrumentedConn wraps a driver.Conn to record query metrics. It must
+// implement every optional driver.Conn interface (QueryerContext,
+// ExecerContext, Pinger, ...) that database/sql knows how to detect, since
+// database/sql picks its execution path based on which interfaces a Conn
+// implements — a wrapper that omits one silently changes behavior (e.g.
+// falling back to a slower path) instead of erroring.
+//
+// Following the pattern used by github.com/XSAM/otelsql (see conn.go
+// there), each method checks at call time whether the wrapped conn
+// actually implements the corresponding interface:
+//   - If not, it returns driver.ErrSkip (the sentinel database/sql defines
+//     for exactly this situation), which tells database/sql to fall back to
+//     the next mechanism (e.g. prepare-then-query) — identical to what
+//     would happen without any wrapping.
+//   - If so, it forwards the call, recording metrics around it where
+//     relevant (QueryContext/ExecContext).
+type instrumentedConn struct {
+	driver.Conn
+	metrics *metrics.Metrics
+}
+
+func wrapConn(conn driver.Conn, m *metrics.Metrics) driver.Conn {
+	return &instrumentedConn{Conn: conn, metrics: m}
+}
+
+var (
+	_ driver.Pinger             = (*instrumentedConn)(nil)
+	_ driver.Queryer            = (*instrumentedConn)(nil) //nolint:staticcheck
+	_ driver.QueryerContext     = (*instrumentedConn)(nil)
+	_ driver.Execer             = (*instrumentedConn)(nil) //nolint:staticcheck
+	_ driver.ExecerContext      = (*instrumentedConn)(nil)
+	_ driver.ConnPrepareContext = (*instrumentedConn)(nil)
+	_ driver.ConnBeginTx        = (*instrumentedConn)(nil)
+	_ driver.SessionResetter    = (*instrumentedConn)(nil)
+	_ driver.NamedValueChecker  = (*instrumentedConn)(nil)
+	_ driver.Validator          = (*instrumentedConn)(nil)
+)
+
+func (c *instrumentedConn) Ping(ctx context.Context) error {
+	pinger, ok := c.Conn.(driver.Pinger)
+	if !ok {
+		return nil
+	}
+	return pinger.Ping(ctx)
+}
+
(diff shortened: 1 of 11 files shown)


internal/database/instrumented.go
```diff
@@ -0,0 +1,214 @@
...
+	return wrapConn(conn, c.metrics), nil
+}
+
+// instrumentedConn wraps a driver.Conn to record query metrics. Unlike an
+// earlier version of this file, it does NOT try to detect which optional
+// driver.Conn interfaces (QueryerContext, ExecerContext, Pinger, ...) the
+// wrapped conn implements and construct a matching wrapper type — that
+// approach requires exhaustively enumerating every combination of ~8
+// optional interfaces, and it's easy to silently drop one (as a test here
+// caught for driver.Pinger).
```
patrick-east: We probably don't need commentary from previous iterations included in the comments. It seems like the only take away that is important here is that `instrumentedConn` must implement all of the `driver.Conn` interfaces.

internal/database/instrumented_test.go
```diff
@@ -0,0 +1,143 @@
...
+		t.Fatalf("QueryContext error = %v, want driver.ErrSkip", err)
+	}
+	if _, err := wrapped.(driver.ExecerContext).ExecContext(context.Background(), "INSERT INTO t VALUES (1)", nil); !errors.Is(err, driver.ErrSkip) {
+		t.Fatalf("ExecContext error = %v, want driver.ErrSkip", err)
+	}
+}
+
+func TestInstrumentedConn_ForwardsUnrelatedOptionalInterfaces(t *testing.T) {
+	// Pinger is never touched by our wrapper; verify it's still forwarded
+	// (the earlier, type-switch-based design silently dropped this).
```
patrick-east: Similar here, we don't need comments about previous issues

patrick-east (commented):
Code changes LGTM, just a couple things I'd recommend cleaning up in the comments

ashutosh-narkar (commented):
The changes look good. We should add some tests to exercise the `metrics` package.

internal/database/instrumented_test.go
yi-chen-roger: updated

internal/database/instrumented.go
yi-chen-roger: updated

pkg/metrics/metrics_test.go
```diff
@@ -87,6 +117,39 @@ func TestGitSyncRecording(t *testing.T) {
...
+
+	if got := testutil.ToFloat64(m.databaseQueryCount.WithLabelValues("insert", "FAILED")); got != 2 {
+		t.Errorf("FAILED count: want 2, got %v", got)
+	}
+	if got := testutil.ToFloat64(m.databaseQueryCount.WithLabelValues("select", "SUCCESS")); got != 1 {
+		t.Errorf("SUCCESS count: want 1, got %v", got)
+	}
+}
+
+func TestDatabaseOperation(t *testing.T) {
```
ashutosh-narkar, line 135: Not sure how this test fits into what we're implementing here. Please remove if not needed.

ashutosh-narkar (commented):
Thanks for adding the tests @yi-chen-roger. The changes lgtm. I'll let @patrick-east give the final sign-off.

pkg/metrics/metrics_test.go
yi-chen-roger: DatabaseOperation is new logic added by this PR (parses the SQL keyword for the operation label) — this test covers its parsing edge cases (unrecognized/empty/multi-keyword input) directly, separate from the end-to-end recording test.

patrick-east (approved):

PR #381, opened 2026-07-28
gitsync/httpsync: add CheckAccess for lightweight source validation

yi-chen-roger:
## Summary
- Adds CheckAccess(ctx) to the git and HTTP/S3 synchronizers (internal/gitsync, internal/httpsync) to verify a source is reachable and its credentials are valid, without performing a full clone/download.
- Adds CheckAccess to the pkg/sync.Synchronizer interface so external callers can validate a source configuration up front instead of waiting for a real sync to fail.
- Git check uses a lightweight ls-remote (via go-git's Remote.List); HTTP uses the configured request against the target without persisting the response; S3 uses HeadObject.

## Test plan
- [x] go build ./...
- [x] go vet ./...
- [x] New unit tests: TestGitsyncCheckAccess, TestHTTPDataSynchronizer_CheckAccess
- [x] Existing test suites unaffected


pkg/sync/interfaces.go
```diff
@@ -18,6 +18,14 @@ type Synchronizer interface {
 	// Returns an error if synchronization fails.
 	Execute(ctx context.Context) error
 
+	// CheckAccess verifies that the configured source is reachable and that its
+	// credentials, if any, are valid, without performing a full synchronization
+	// (e.g. no git clone or full download). It allows callers to validate a
+	// source configuration up front rather than waiting for a real sync to fail.
+	//
+	// Returns an error if the source is unreachable or the credentials are invalid.
+	CheckAccess(ctx context.Context) error
```
srenatus, line 37: Q: do we currently allow external implementations? I don't think so, but please check: if we do, this should be an interface to opt-in to.

srenatus (approved):
Correct me if I'm wrong, please: This is added as a method that an OCP-embedding service can use, it's not wired up in any way with health checking, or other pre-existing logic in OCP (run as server). If so, let's get this merged!

pkg/sync/interfaces.go
yi-chen-roger: separated the interfaces

internal/httpsync/httpsync.go
```diff
@@ -181,6 +181,69 @@ func (*HttpDataSynchronizer) Close(context.Context) {
 	// No resources to close for HTTP synchronizer
 }
 
+// CheckAccess verifies that the configured HTTP/S3 endpoint is reachable and
+// that the configured credentials, if any, are valid. It always issues a HEAD
+// request, regardless of the configured method/body, so that it never sends
+// the real request payload or triggers side effects of a non-idempotent sync.
+func (s *HttpDataSynchronizer) CheckAccess(ctx context.Context) error {
+	if s.region != "" {
```
ashutosh-narkar, line 189: Is there some precedent for this implicit behavior?

yi-chen-roger: Good question — no established precedent in this codebase (this is the first CheckAccess on the interface). The reasoning for hardcoding HEAD instead of the configured method/body:

  - CheckAccess is meant to be a read-only, side-effect-free probe — it can run frequently (e.g. on every source-config save) and shouldn't risk triggering the real endpoint's side effects if the configured method is POST/PUT with a
  body.
  - HEAD also avoids downloading the actual response payload, keeping the check lightweight regardless of how large the real sync response would be.

pkg/sync/interfaces.go
srenatus: Does that mean the answer is „yes“? Hmm. I didn’t expect that. Let’s not do this if there’s no reason to.

ashutosh-narkar (approved):

PR #382, opened 2026-07-31
pkg/service: add ValidateSourceAccess for lightweight source validation

yi-chen-roger:
## Summary
- Adds `service.ValidateSourceAccess(ctx, src, provider)`, a standalone function that checks whether a `Source`'s git and datasource (http/s3) bindings are reachable and their credentials, if any, valid — without requiring `Service.Init` or a database connection.
- Builds on the `CheckAccess` support added to gitsync/httpsync in #381: constructs the same internal synchronizer types used during a real sync and calls `CheckAccess` on each configured binding.
- Each `BindingAccessResult` also reports `UserError`, derived from `syncerr.IsUserError`, so callers can distinguish misconfiguration (invalid credentials, unreachable repo/URL, 4xx response) from a transient/internal failure.
- This closes the gap left by #381: `gitsync`/`httpsync`/`syncerr` are internal packages, so external callers had no way to construct a `Synchronizer` and call `CheckAccess` themselves, even though `pkg/sync.AccessChecker` is public.

## Test plan
- [x] go build ./...
- [x] go vet ./...
- [x] New unit tests: `TestValidateSourceAccess_NoBindings`, `TestValidateSourceAccess_Git`, `TestValidateSourceAccess_HTTPDatasource`, `TestValidateSourceAccess_UnknownDatasourceTypeSkipped`
- [x] Existing test suites unaffected

### DIFF
--- a/pkg/service/validate.go
+++ b/pkg/service/validate.go
@@ -11,20 +11,43 @@ import (
 	pkgsync "github.com/open-policy-agent/opa-control-plane/pkg/sync"
 )
 
+// BindingType identifies the kind of binding a BindingAccessResult reports on.
+type BindingType string
+
+const (
+	BindingTypeGit  BindingType = "git"
+	BindingTypeHTTP BindingType = "http"
+	BindingTypeS3   BindingType = "s3"
+)
+
+// Error is a JSON-serializable description of why a binding's access check
+// failed. A Go error doesn't marshal to anything useful on its own, so
+// BindingAccessResult reports failures through this type instead.
+type Error struct {
+	Message string `json:"message"`
+	// UserError is true when the failure is caused by misconfiguration
+	// (invalid credentials, an unreachable repository/URL, a 4xx
+	// response, ...) rather than a transient or internal failure. Callers
+	// can use this to decide whether Message is safe to surface to the
+	// end user as-is.
+	UserError bool `json:"user_error"`
+}
+
+func newError(err error) *Error {
+	if err == nil {
+		return nil
+	}
+	return &Error{Message: err.Error(), UserError: syncerr.IsUserError(err)}
+}
+
 // BindingAccessResult reports the outcome of checking one git or datasource
 // binding within a Source for reachability and credential validity.
 type BindingAccessResult struct {
-	// Type is "git", "http", or "s3".
-	Type string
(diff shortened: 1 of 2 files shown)


pkg/service/validate.go
```diff
@@ -0,0 +1,98 @@
...
+	// Name is the datasource name, or "" for the source's git binding.
+	Name string
+	// Err is non-nil if the binding could not be verified.
+	Err error
+	// UserError is true when Err is caused by misconfiguration (invalid
+	// credentials, an unreachable repository/URL, a 4xx response, ...)
+	// rather than a transient or internal failure. Callers can use this to
+	// decide whether Err is safe to surface to the end user as-is.
+	UserError bool
+}
```
ashutosh-narkar: An error type would be helpful. For example:

```go
type Error struct {
	Message  string        `json:"message"`
}
```

pkg/service/validate.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/open-policy-agent/opa-control-plane/internal/httpsync"
+	"github.com/open-policy-agent/opa-control-plane/internal/syncerr"
+	pkgsync "github.com/open-policy-agent/opa-control-plane/pkg/sync"
+)
+
+// BindingAccessResult reports the outcome of checking one git or datasource
+// binding within a Source for reachability and credential validity.
+type BindingAccessResult struct {
+	// Type is "git", "http", or "s3".
+	Type string
```
ashutosh-narkar: What about for a type for this too?

ashutosh-narkar (commented):
Changes look fine. Few small comments.

pkg/service/validate.go
yi-chen-roger: Done in 7735233 — added a `service.Error{Message, UserError}` type (JSON-serializable) and `BindingAccessResult.Err` is now `*service.Error`.

pkg/service/validate.go
yi-chen-roger: Done in 7735233 — added a `service.BindingType` string type with `BindingTypeGit`/`BindingTypeHTTP`/`BindingTypeS3` constants.

ashutosh-narkar (approved):

PR #385, opened 2026-08-03
Track when bundle status was last updated

ashutosh-narkar:
Currently OCP does not record any timing for status updates. Callers might find this timing information useful.

This change adds an updated_at column to bundles_statuses that is written on every upsert, and exposes it through the BundleStatus API type so callers can read it.

### DIFF
--- a/internal/database/bundle_status.go
+++ b/internal/database/bundle_status.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/open-policy-agent/opa-control-plane/pkg/config"
 )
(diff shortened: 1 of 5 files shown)


internal/database/bundle_status.go
```diff
@@ -39,8 +40,12 @@ func (d *Database) UpsertBundleStatus(ctx context.Context, tenant, bundle, revis
 			return fmt.Errorf("error looking up bundle %s: %w", bundle, err)
 		}
 
-		resID, err := d.upsertReturning(ctx, true, true, tx, tenant, "bundles_statuses", []string{"bundle_id", "revision", "phase", "status", "error_message"}, []string{"bundle_id", "revision"},
-			bundleID, revision, phase, status, errMsg)
+		// updated_at is written on every upsert so callers
```
yi-chen-roger, line 43: nit: this comment sits right above the time.Now() line rather than above the whole upsert call

yi-chen-roger (approved):
LGTM

PR #386, opened 2026-08-03
service: fix zero-value BuildState being misreported as INTERNAL_ERROR

yi-chen-roger:
## Summary
- `BuildStateInternalError` was defined as `BuildState`'s zero value, so any `BundleWorker` whose status was never populated via `report()` (e.g. it exits early through `die()` on a configuration change or shutdown before completing a build iteration) was misreported as `INTERNAL_ERROR` with an empty error message, even though no real error occurred.
- Adds `BuildStateUnknown` as the explicit zero value/state, and has `die()` mark the worker's status accordingly whenever it exits without ever calling `report()`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/service/...` (excluding pre-existing network-bound tests unrelated to this change)

### DIFF
--- a/pkg/service/worker.go
+++ b/pkg/service/worker.go
@@ -43,6 +43,7 @@ type BundleWorker struct {
 	log           *logging.Logger
 	bar           *progress.Bar
 	status        Status
+	reported      bool
 	interval      time.Duration
 	database      *database.Database
 	tenant        string
@@ -256,20 +257,13 @@ func (w *BundleWorker) Execute(ctx context.Context) time.Time {
 
 func (w *BundleWorker) report(ctx context.Context, state BuildState, phase BuildPhase, revision string, startTime time.Time, err error) time.Time {
 	interval := w.interval
-	w.status.State = state
 	msg := ""
 	if err != nil {
 		interval = errorInterval // faster retry on error
 		msg = err.Error()
-		w.status.Message = msg
 	}
 
-	if w.database != nil {
-		if _, uerr := w.database.UpsertBundleStatus(ctx, w.tenant, w.bundleConfig.Name,
-			revision, phase.String(), state.String(), msg); uerr != nil {
-			w.log.Warnf("failed to track bundle status %q: %v", w.bundleConfig.Name, uerr)
-		}
-	}
+	w.writeStatus(ctx, state, phase, revision, msg)
 
 	if state == BuildStateSuccess {
 		w.metrics.BundleBuildSucceeded(w.bundleConfig.Name, state.String(), startTime)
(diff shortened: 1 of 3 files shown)


patrick-east (commented):
Having the default as an unknown seems like a good improvement, but for the issue described should that be more like a "cancelled" state? Or we didn't know the reason why it didn't reach a terminal state?

patrick-east (commented):
Can we get any test coverage for the early exit state?

yi-chen-roger:
> Can we get any test coverage for the early exit state?

test added

yi-chen-roger:
> Having the default as an unknown seems like a good improvement, but for the issue described should that be more like a "cancelled" state? Or we didn't know the reason why it didn't reach a terminal state?

updated

pkg/service/worker.go
```diff
@@ -302,6 +304,10 @@ func (w *BundleWorker) configurationChanged() bool {
 }
 
 func (w *BundleWorker) die(ctx context.Context) time.Time {
+	if !w.reported {
+		w.status = Status{State: BuildStateCancelled, Message: "worker stopped due to a configuration change or shutdown before completing a build iteration"}
```
patrick-east: I haven't looked very closely at the flow for this, so maybe a dumb question, but would it be worth trying to do a `report()` when it dies to mark the status? Maybe better question, what happens to the bundle status in the DB for one of the builds that would hit this condition?

yi-chen-roger: Good question — i thought about it, but decided not to call report() directly from die() for now:

1. Recursion risk: report() itself calls w.die(ctx) at the end when singleShot is true. If die() also called report(), we'd get die() → report() → die(), which double-closes w.done and panics.
 2. Semantic mismatch: report() also drives errorInterval (faster retry) and metrics.BundleBuildFailed/Succeeded. A worker being torn down due to config change/shutdown isn't a failed build attempt, so piggybacking on those side effects doesn't seem right.
You're correct that this means the DB status row for the bundle isn't updated when a worker dies before its first report() — it'll just retain whatever the previous iteration wrote.  That's a real gap, but a separate one from what this PR fixes，will do that in a separate PR

patrick-east: > Recursion risk: report() itself calls w.die(ctx) at the end when singleShot is true. If die() also called report(), we'd get die() → report() → die(), which double-closes w.done and panics.

Right, we'd need to have some protection for that, or refactor a little bit to avoid the recursion. The part I would mostly be interested in is the DB insert for the status.

> Semantic mismatch: report() also drives errorInterval (faster retry) and metrics.BundleBuildFailed/Succeeded. A worker being torn down due to config change/shutdown isn't a failed build attempt, so piggybacking on those side effects doesn't seem right.

Not sure I'm following this part. Why would the interval matter if it is shutting down? I'm assuming we'd pretty much ignore whatever interval gets returned.

> You're correct that this means the DB status row for the bundle isn't updated when a worker dies before its first report() — it'll just retain whatever the previous iteration wrote.

What would be retained? Like as a user of this what feedback do I have about this build?

yi-chen-roger: - DB insert: Now handled — report() and die() share a writeStatus() helper. 
- "What would be retained": That gap is now closed — the user sees a CANCELLED status instead of the previous iteration's stale status (or no row at all).

patrick-east (approved):
Thanks Yi!

PR #387, opened 2026-08-03
metrics: add git_sync_server_error_total for non-user git failures

ashutosh-narkar:
The git_sync_count_total{state="FAILED"} counter increments on every git
sync failure, regardless of cause. This conflates two very different
classes of error: user misconfiguration (a bad repo URL, invalid
credentials etc.) versus server-side or transient
failures (network issues, remote outages etc.).

This change adds a dedicated git_sync_server_error_total counter
that records only non-user failures. This can help operators to
alert on git_sync_server_error_total to catch real sync problems.

### DIFF
--- a/internal/gitsync/gitsync.go
+++ b/internal/gitsync/gitsync.go
@@ -142,6 +142,7 @@ func (s *Synchronizer) Execute(ctx context.Context) (map[string]any, error) {
 		if isGitUserError(err) {
 			return nil, syncerr.UserError{Cause: wrapped}
 		}
+		s.metrics.GitSyncServerError(s.sourceName, s.config.Repo)
 		return nil, wrapped
 	}
 	if done {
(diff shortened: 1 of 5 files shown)


yi-chen-roger:
The PR description explicitly calls out "a missing reference" as an example of user misconfiguration that should be excluded from the new server-error counter. But isGitUserError (unchanged by this PR)   only matches:
 transport.ErrRepositoryNotFound, ErrAuthorizationFailed, ErrAuthenticationRequired, ErrEmptyRemoteRepository

It does not match a missing/nonexistent branch/tag reference

ashutosh-narkar:
> The PR description explicitly calls out "a missing reference" as an example of user misconfiguration that should be excluded from the new server-error counter. But isGitUserError (unchanged by this PR) only matches: transport.ErrRepositoryNotFound, ErrAuthorizationFailed, ErrAuthenticationRequired, ErrEmptyRemoteRepository
> 
> It does not match a missing/nonexistent branch/tag reference

Updated the commit description.

yi-chen-roger (approved):
lgtm

PR #394, opened 2026-08-14
fix(builder): keep VCS metadata out of the bundle revision hash

ashutosh-narkar:
Bundle revisions resolved from `input.bundle.hash` changed for identical content. A new version on every re-clone was created as the hash was computed over the whole checkout, `.git`(specifically index) included.

### DIFF
--- a/internal/gitsync/gitsync.go
+++ b/internal/gitsync/gitsync.go
@@ -41,6 +41,15 @@ import (
 // bundle builder.
 const configFile = "ocpconfig"
 
+// MetadataFiles are glob patterns matching git's own bookkeeping inside a
+// checkout.
+var MetadataFiles = []string{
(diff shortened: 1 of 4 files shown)


srenatus:
Hah, good point. I wonder if we could reuse the go-git library for this, it might offer an FS walk method that also skips files according to `.gitignore` 🤔

ashutosh-narkar:
> Hah, good point. I wonder if we could reuse the go-git library for this, it might offer an FS walk method that also skips files according to `.gitignore` 🤔

I wasn't able to find something that would fit in well. Happy to incorporate if you have some ideas. FWIW I've refactored the code to move the git patterns in the `gitsync` package.

srenatus (approved):

PR #399, opened 2026-08-17
service: skip build when resolved revision already matches stored bundle

yi-chen-roger:
## Summary

Today, `Upload()` implementations can already skip re-uploading identical bundle content via `ErrNotModified`, but that check only runs after the full sync/transform/build pipeline has already produced the bundle bytes to hash. For bundles whose revision doesn't depend on the built bundle's own content hash (i.e. doesn't reference `input.bundle.hash`), the revision is fully resolvable right after sync, before transform/build ever run.

This PR adds an optional `VersionedObjectStorage` interface with a single `LatestRevision` method. `BundleWorker.Execute` now resolves the revision early in that case and, if a storage backend implements the new interface and reports the same revision is already the latest stored one, skips the rest of the iteration entirely instead of rebuilding from scratch every poll.

`LatestRevision` is implemented for the S3 and GCS backends by reusing their existing metadata-only `HeadObject`/`Attrs` calls (no download required) against the revision they already record on `Upload`. Azure and the local filesystem backend don't currently record a revision this way, so they're left as-is — callers on those backends simply fall back to the existing post-build check (since they don't implement the new optional interface).

## Why draft

Opening as draft to get early feedback on the interface shape (optional capability interface + type assertion vs. adding directly to `ObjectStorage`) before polishing tests/docs further.

## Test plan
- `go build ./...` — clean
- `go vet ./...` — clean
- `golangci-lint run ./internal/s3/... ./pkg/objectstorage/... ./pkg/service/...` — 0 issues
- `go test ./pkg/service/... -run 'TestRevision|TestExtractRevisionRefs|TestResolveRevision|TestBundleWorker'` — passing
- Note: `TestService` and `TestS3` (the httptest-based e2e/integration tests) couldn't be run in my current sandbox (`bind: operation not permitted` for local test listeners, pre-existing/unrelated to this change — confirmed by reproducing the same failure on a clean checkout of `main`). Would appreciate CI running these.
- [ ] Add dedicated unit tests for the new early-skip path in `worker.go` (not yet added — will follow up once the interface shape is agreed on)

### DIFF
--- a/internal/s3/s3_test.go
+++ b/internal/s3/s3_test.go
@@ -7,11 +7,15 @@ import (
 	"encoding/hex"
 	"errors"
 	"io"
+	"net/http"
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strconv"
 	"testing"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/fsouza/fake-gcs-server/fakestorage"
 	"github.com/johannesboyne/gofakes3"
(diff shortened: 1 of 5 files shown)


internal/s3/s3_test.go
```diff
@@ -265,6 +286,62 @@ func TestS3NotModified(t *testing.T) {
...
 
+func TestS3LatestRevision(t *testing.T) {
+	// Dispatches directly to gofakes3's handler in-process, so this test
+	// doesn't need to bind a real TCP listener.
+	mock := s3mem.New()
+	if err := mock.CreateBucket("test"); err != nil {
+		t.Fatal(err)
+	}
+
+	ctx := context.Background()
```
srenatus: `t.Context()` please

yi-chen-roger:
Closing — the underlying revision instability (VCS metadata leaking into the bundle content hash) was the real problem, and that's now fixed by #394. Not pursuing this early-skip optimization further for now.

PR #402, opened 2026-08-28
fix(builder): keep root set minimal in refSet.add

### LINKED ISSUE #279: manifest that opactl generates inside the bundle misses some roots from the bundle
Hello,

I build my bundle with ocp and create a local archive. When I extract the archive I checked the manifest and it does not contain many of the roots that I have in my policies. If I try to start `opa` with that bundle, it tells me: 

```
error: load error: bundle /bundles/mybundle.tar.gz: manifest roots [ roots ... ] ] do not permit 'package some.package in module '/dir/jpolicies/my_policy.rego'
```

If I run `opa` with `--ignore=.*`, then opa can start.

CD57:
Manifests built by OCP could miss roots for packages present in the bundle, so OPA refuses to load it: 
"error:` bundle bundle.tar.gz: manifest roots [alpha] do not permit 'package gamma' in module '/example-policies/gamma/p.rego'"

refSet.add in pkg/builder/builder.go had two bugs. First, the loop over its args returned on the first ref that overlapped an existing entry, losing every ref after it, which is what drops gamma in the [issue](https://github.com/open-policy-agent/opa-control-plane/issues/279#issuecomment-4746260421). And second, adding a parent of several existing refs replaced only the first child, leaving overlapping roots (i.e. "manifest has overlapped roots: 'alpha' and 'alpha/c2'"), which can happen when there is more than one filesystem, e.g. directory plus git.

This change rewrites the add function to skip refs covered by an existing entry and drop every entry a new ref covers. Replacing the returns with continues (as suggested in the issue thread) would fix the first bug but not the second. Added a regression test per bug. Both fail on the old code with the errors above.

I was looking into this issue a few months ago and totally forgot about it. Appreciate any feedback/review, thank you.

Fixes #279

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -509,18 +509,30 @@ type refSet struct {
 	refs []ast.Ref
 }
 
+// add inserts each ref into the set while keeping it minimal: no ref in the
+// set is a prefix of another. Adding a ref already covered by (a prefix of) an
+// existing ref is a no-op; adding a ref that is a prefix of existing refs
+// replaces all of them.
 func (rs *refSet) add(ns ...ast.Ref) {
+next:
 	for _, n := range ns {
-		for i, r := range rs.refs {
-			switch {
-			case r.HasPrefix(n):
-				rs.refs[i] = n
-				return
-			case n.HasPrefix(r):
-				return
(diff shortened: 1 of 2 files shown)


srenatus (approved):
Thank you! Good find. 👏

PR #405, opened 2026-09-02
database: propagate the caller's context into every query

yi-chen-roger:
## Summary

`sql.Tx.Query`, `.Exec` and friends without the `Context` suffix pass `context.Background()` to the driver — [`database/sql/sql.go`](https://github.com/golang/go/blob/master/src/database/sql/sql.go):

```go
func (tx *Tx) Query(query string, args ...any) (*Rows, error) {
	return tx.QueryContext(context.Background(), query, args...)
}
```

Eight queries in `internal/database` use those variants, so they run detached from the caller they belong to, even though each is already inside a method that takes `ctx`.

## Why it matters

- **Cancellation and deadlines don't reach the database.** A client that disconnects, or a request that hits its deadline, leaves these queries running to completion. `ListBundles`/`ListSources` are wide multi-table joins with row fan-out, so the abandoned work is not cheap.
- **A connection-acquisition timeout applied to the caller's context is bypassed** for these queries.
- **Tracing breaks in a way that is actively misleading rather than merely absent.** A span opened around one of these queries has no parent in context, so it's exported as the root of its own trace instead of nesting under the request. A trace of a `GetBundle` call showed only the surrounding transaction's `BEGIN`/`SAVEPOINT`/`RELEASE`/`COMMIT` spans, with a half-second gap where the `SELECT` should have been — the statement *was* traced, just filed under a trace of its own, which is how this was found.

## What changed

Eight call sites switched to the `Context`-taking variants. No behaviour change beyond threading through the `ctx` that was already in scope:

| Location | Method |
|---|---|
| `database.go` | `sourcesDataGet`, `SourcesDataDelete`, `ListBundles`, `ListSources` (main query), `ListSources` (datasources), `ListSecrets`, `ListStacks` |
| `bundle_status.go` | `ListBundleStatuses` |

After this change `grep` finds no remaining non-`Context` `Query`/`Exec`/`QueryRow`/`Prepare` calls in `internal/database`.

## Test plan

- `go build ./...` — clean
- `go vet ./internal/database/...` — clean
- `gofmt -l` — clean
- `go test ./internal/database/...` — the SQLite-backed subtests all pass (72 subtests of `TestDatabase`, 0 failures). The testcontainer-backed dialects (postgres/mysql/cockroachdb) couldn't run in my environment — Docker isn't available, and every failure is a container-startup error, with the failure count matching the "rootless Docker not found" count exactly. Would appreciate CI running those.

### DIFF
--- a/internal/database/bundle_status.go
+++ b/internal/database/bundle_status.go
@@ -202,7 +202,7 @@ func (d *Database) ListBundleStatuses(ctx context.Context, principal, tenant, bu
 			args = []any{tenant, bundleID, limit}
 		}
 
-		rows, err := tx.Query(query, args...)
+		rows, err := tx.QueryContext(ctx, query, args...)
 		if err != nil {
 			return err
 		}
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -529,7 +529,7 @@ func sourcesDataGet[T any](ctx context.Context, d *Database, sourceName, path st
 			return zero, false, fmt.Errorf("lookup source name %s: %w", sourceName, err)
 		}
 
-		rows, err := tx.Query(fmt.Sprintf(`SELECT
+		rows, err := tx.QueryContext(ctx, fmt.Sprintf(`SELECT
 	data
 FROM
 	sources_data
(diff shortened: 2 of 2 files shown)


srenatus (approved):
Yay thanks

PR #417, opened 2026-09-10
fix: data sources with same name but different paths overriding data

### LINKED ISSUE #416: Data sources with the same `name` but different `path` silently overwrite each other
## Summary
A Source with two data sources sharing the same `name` but different `path` values loses one of them on sync — silently, with no error.

## Steps to Reproduce
1. Configure a Source with two data sources:
   - `name: duplicate`, `type: http`, no `path`
   - `name: duplicate`, `type: http`, `path: subpath`
2. Sync the config and let the bundle build.
3. Inspect the built bundle's `data.json`.

**Expected:** both appear (`data.duplicate` and `data.subpath.duplicate`).
**Actual:** only one survives (whichever was upserted last); the other is gone with no error.

## Suspected Root Cause
`internal/database/database.go`, `UpsertSource`:
```go
d.upsertRel(ctx, tx, "sources_datasources",
    []string{"source_id", "name", "type", "path", "config", "transform_query", "secret_id", "credentials_name"},
    []string{"source_id", "name"},   // conflict key excludes "path"
    ...)
```
This becomes `ON CONFLICT (source_id, name) DO UPDATE ...`. The table schema (`internal/migrations/migrations_next.go`, `migrations_v0.2.0.go`) matches: `PrimaryKey("source_id", "name")`.

Since `path` isn't part of the key, two datasources with the same `name` collapse into one row — the later upsert overwrites `type`, `path`, `config`, etc. of the earlier one. This contradicts `config.Datasource` (`pkg/config/config.go`), where `Name` and `Path` are independent fields with nothing preventing multiple entries sharing a `Name`.

## Impact
Silent data loss of configured data sources → incomplete policy data with no diagnostic signal.

M4nuF:
## Fix: data sources with the same `name` but different `path` silently overwrite each other

Fixes #416

### Changes
- Widen `sources_datasources` primary key to `(source_id, name, path)` via new migration (recreate table across sqlite/postgres/mysql/cockroachdb).
- Update `UpsertSource`'s conflict key to `(source_id, name, path)`.
- Fix `Datasources.Equal` to key by `(Name, Path)` instead of `Name` alone.
- Add regression test covering two datasources with the same name but different paths.

### Testing
- `go test ./...` (sqlite, postgres, mysql, cockroachdb via testcontainers) — all pass.

### DIFF
--- a/internal/migrations/migrations_next.go
+++ b/internal/migrations/migrations_next.go
@@ -120,6 +120,53 @@ func addBundlesStatusesUpdatedAt(offset int, dialect string) fs.FS {
 	})
 }
 
+// fixDatasourcesPrimaryKey widens the primary key of sources_datasources from
+// (source_id, name) to (source_id, name, path). Without "path" in the key,
+// two datasources sharing the same "name" but different "path" values collapse
+// into a single row on upsert, silently dropping one of them. Since none of our
+// dialects support altering a primary key in place, we recreate the table with
+// the new key, copy the data across, and drop the old table. The four steps are
+// split into separate migrations (rather than one multi-statement transaction)
+// because CockroachDB's async schema changer rejects DML against a table that
+// was just created earlier in the same transaction.
+func fixDatasourcesPrimaryKey(offset int, dialect string) fs.FS {
+	var kind int
+	switch dialect {
+	case "postgresql":
+		kind = postgres
+	case "mysql":
+		kind = mysql
+	case "sqlite":
+		kind = sqlite
+	case "cockroachdb":
+		kind = cockroachdb
+	}
+
+	newTbl := createSQLTable("sources_datasources").
+		WithIteration("ocp_v2_pkfix"). // distinct from "ocp_v2" to avoid clashing with the old table's constraint names while both exist during the migration
+		VarCharNonNullColumn("name").
+		IntegerNonNullColumn("source_id").
+		IntegerColumn("secret_id"). // optional
+		TextNonNullColumn("type").
+		VarCharNonNullColumn("path"). // part of the primary key; needs a bounded length for MySQL
+		TextNonNullColumn("config").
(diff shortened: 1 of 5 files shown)


internal/migrations/migrations_next.go
```diff
@@ -120,6 +120,53 @@ func addBundlesStatusesUpdatedAt(offset int, dialect string) fs.FS {
...
+		ForeignKey("source_id", "sources(id)")
+
+	const oldName = "sources_datasources_old"
+	cols := "name, source_id, secret_id, type, path, config, transform_query, credentials_name"
+
+	return ocp_fs.MapFS(map[string]string{
+		fmt.Sprintf("%03d_fix_datasources_primary_key_rename.up.sql", offset):   "ALTER TABLE sources_datasources RENAME TO " + oldName,
+		fmt.Sprintf("%03d_fix_datasources_primary_key_create.up.sql", offset+1): strings.TrimRight(newTbl.SQL(kind), ";"),
+		fmt.Sprintf("%03d_fix_datasources_primary_key_copy.up.sql", offset+2):   fmt.Sprintf(`INSERT INTO sources_datasources (%[1]s) SELECT %[1]s FROM %[2]s`, cols, oldName),
+		fmt.Sprintf("%03d_fix_datasources_primary_key_drop.up.sql", offset+3):   "DROP TABLE " + oldName,
```
srenatus, line 166: I think we have one txn per file -- can we put this into a single transaction to avoid inconsistencies on failure? 🤔

srenatus:
Heya. Thanks for working on a fix. Easy things first: DCO needs to be adjusted on the commits.

Harder things: there might be other places where a data source is referenced by name only -- like https://github.com/open-policy-agent/opa-control-plane/blob/main/pkg/service/worker.go#L174-L183. Can you add some sort of e2e test so we know that it'll actually work? 💦 

Thanks again 🎈

PR #418, opened 2026-09-10
database: fold the existence check into the id lookup

yi-chen-roger:
## Why

Seven code paths ran **two statements with identical WHERE clauses**: an existence check (`SELECT 1 …`) followed by a lookup for the same row's id (`SELECT id …`).

```sql
-- resourceExists
SELECT 1  FROM bundles WHERE name = $1 AND tenant_id = (SELECT id FROM tenants WHERE name = $2)
-- lookupID, immediately after
SELECT id FROM bundles WHERE name = $1 AND tenant_id = (SELECT id FROM tenants WHERE name = $2)
```

The second answers the first. In a production trace of `DeleteBundleConfig` those two statements cost **457ms and 324ms** against the same row — a full round trip spent learning something the next statement was about to return anyway.

## What changed

`lookupExistingID` reports a missing row as `ErrNotFound`, so one statement answers both questions and callers return the same errors as before. It sits alongside the existing `lookupRequiredID`, which translates to `ErrInvalidReference`.

Seven call sites lose a statement:

| | how it deduplicated |
|---|---|
| `DeleteBundle`, `DeleteSource`, `DeleteSecret`, `DeleteStack` | `prepareDelete` now returns the id |
| `sourcesDataGet`, `sourcesDataPut`, `SourcesDataDelete` | inline `resourceExists` dropped |

`prepareDelete` checks authorization **before** looking anything up, which preserves the property its old comment described (`// only inform about existence if authorized`) — an unauthorized caller gets `ErrNotAuthorized` and can't distinguish a missing row from a forbidden one.

`resourceExists` keeps its one remaining caller, `prepareUpsert`, which needs the answer to choose between create and update permissions rather than to find an id.

## A small authorization fix falls out of this

`sourcesDataPut` authorizes with a boolean `Check`, but ran the existence check **first**:

```go
if err := d.resourceExists(...); err != nil {   // ErrNotFound leaks here
    return err
}
ad := ...
if !d.authorizer.Check(ctx, tx, d.arg, ad) {
    return errors.New("unauthorized")
}
```

So an unauthorized caller could tell whether a source existed by comparing `ErrNotFound` against `unauthorized`. The surviving lookup now sits after the check.

`sourcesDataGet` and `SourcesDataDelete` authorize with `Partial`, which folds conditions into the query instead of rejecting, so ordering carries no such signal there.

## Testing

- [x] `go build ./...`, `go vet ./internal/database/...`, `gofmt` clean
- [x] `TestDatabase/sqlite-memory-only` and `/sqlite-persistence` pass — these drive all four `Delete*` paths and the `sources_data` ones
- [x] Added a `lookupExistingID` case to the error-translation test in `errors_test.go`, asserting `ErrNotFound` and that `sql.ErrNoRows` doesn't leak through
- [ ] Postgres / MySQL / CockroachDB — **CI only**, see Notes

## Notes

⚠️ **I could not run the non-SQLite tests locally** (testcontainers needs Docker, which isn't available in my environment). The error-translation behaviour is dialect-independent, but please don't merge on a green SQLite run alone.

⚠️ **Error-message change.** Where the lookup fails for a reason other than a missing row, the message is now `lookup bundles my-bundle: …` from `prepareDelete` rather than `lookup bundle my-bundle: …` from the call site — the resource name comes from the table now. Sentinel errors (`ErrNotFound`, `ErrNotAuthorized`) are unchanged, so API status codes are unaffected.

⚠️ **Not the main cost in these requests.** This removes one round trip per delete. The larger items in the same traces are elsewhere: the cascade from `DELETE FROM bundles` into `bundles_statuses`, which has no usable index (#407), and the `ListSources` wide join, which full-scans three tables. Worth landing on its own merits — it's strictly less work for the same result — but it won't move the endpoint's p99 by itself.

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -512,10 +512,6 @@ func sourcesDataGet[T any](ctx context.Context, d *Database, sourceName, path st
 ) func(*sql.Tx) (T, bool, error) {
 	return func(tx *sql.Tx) (T, bool, error) {
 		var zero T
-		if err := d.resourceExists(ctx, tx, tenant, "sources", sourceName); err != nil {
-			return zero, false, err
-		}
-
 		ad := d.accessFactory().WithPrincipal(principal).WithTenant(tenant).WithResource("sources").WithPermission("sources.data.read").WithName(sourceName)
 		expr, err := d.authorizer.Partial(ctx, ad, nil)
 		if err != nil {
@@ -524,9 +520,9 @@ func sourcesDataGet[T any](ctx context.Context, d *Database, sourceName, path st
 
 		conditions, args := expr.SQL(d.arg, []any{sourceName, path})
 
-		args[0], err = d.lookupID(ctx, tx, tenant, "sources", sourceName)
+		args[0], err = d.lookupExistingID(ctx, tx, tenant, "sources", sourceName)
 		if err != nil {
-			return zero, false, fmt.Errorf("lookup source name %s: %w", sourceName, err)
+			return zero, false, err
 		}
 
 		rows, err := tx.QueryContext(ctx, fmt.Sprintf(`SELECT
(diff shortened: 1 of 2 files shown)


srenatus:
Please run the tests on your machine, Claude’s sandbox isn’t there if you try it yourself. Thanks!

PR #423, opened 2026-09-17
database: scope search_path to a configured schema per transaction

yi-chen-roger:
## Fix

OCP's queries use unqualified table names and resolve via `search_path`. When a consumer shares OCP's connection pool and lists another schema first (e.g. `oma,ocp`), every OCP query first does a doomed lookup in that schema. CockroachDB caches positive descriptor lookups but not negative ones, so that miss re-runs on every statement and, when the descriptor leaseholder is remote, costs a cross-region round-trip each time — inflating planning latency.

## Changes

- Add `Database.WithSchema(schema)`: when set, each transaction runs `SET LOCAL search_path = <schema>` before its queries, so OCP's unqualified names resolve in that schema.
- `SET LOCAL` scopes it to the transaction, so it never leaks to other users of a shared connection pool. No-op when no schema is set (the default, preserving existing deployments) or outside PostgreSQL/CockroachDB.
- `WithSchema` validates the schema as a plain identifier and returns an error, so a misconfigured schema fails at setup rather than on every query.
- Route the two reads that ran outside a transaction (`QuerySourceID`, `GetPrincipalID`) through a transaction so they pick up the same `search_path`.

## Testing

- `go test ./...` (sqlite, postgres, mysql, cockroachdb via testcontainers) — all pass.
- New unit tests: `TestWithSchema` (identifier validation) and `TestSearchPathStmt` (per-dialect statement; no-op on MySQL/SQLite).


internal/database/database.go
```diff
@@ -156,6 +158,23 @@ func (d *Database) WithRawRootConfig(rawRootConfig []byte) *Database {
 	return d
 }
 
+// schemaIdent bounds a configured schema to a plain SQL identifier so it can be
+// interpolated into search_path safely.
+var schemaIdent = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
```
srenatus: [nit] The identifier isn't quoted when interpolated ("SET LOCAL search_path = " + d.schema), and the regex (^[A-Za-z_][A-Za-z0-9_]*$) blocks injection, but it still allows mixed-case names — and unquoted identifiers get folded to lowercase by Postgres/CRDB. So WithSchema("MySchema") would silently resolve to myschema, which could be a silent misconfiguration if anyone ever passes a mixed-case schema. We could either lowercase-normalize in WithSchema, restrict the regex to lowercase, or quote the identifier (fmt.Sprintf("SET LOCAL search_path = %q", d.schema), using double quotes not Go's %q, so pq.QuoteIdentifier-style).

srenatus (approved):
one nitpick, LGTM otherwise

internal/database/database.go
yi-chen-roger: Thanks for pointing out that.
I will restrict the regex to lowercase

